PR #6, opened 2024-07-04
build: simplify Docker builds

wpf500:
This is my attempt to understand and simplify the new Docker builds in the monorepo. It also fixes the problem that the `.dockerignore` files were being ignored (haha), as it needs to be in the root of the context.

Mostly it reduces the number of lines needed to add the common package and removes as much context specific knowledge as possible.

### DIFF
--- a/apps/frontend/vite.config.ts
+++ b/apps/frontend/vite.config.ts
@@ -9,7 +9,6 @@ import theme from './plugins/theme';
 
 export default ({ command, mode }) => {
   const env = loadEnv(mode, process.cwd(), '');
-  console.debug('API_PROXY_URL', env.API_PROXY_URL);
 
   const plugins = [
     vue(),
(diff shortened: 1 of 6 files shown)


apps/frontend/vite.config.ts
```diff
@@ -47,6 +46,8 @@ export default ({ command, mode }) => {
 
   // Use environment variables when developing locally
   if (command === 'serve') {
+    console.debug('API_PROXY_URL', env.API_PROXY_URL);
```
wpf500, line 49: This change is just to stop it printing `API_PROXY_URL undefined` on build, which could be perceived as an error but isn't.

JumpLink: Why does entering a log print prevent the output of a log print?

JumpLink:
@wpf500 It's good to know that the `.dockerignore` simply has to be placed in the root / where the context is. It can then be kept general enough to be used for all docker images. To do this, it would make sense to align the structure of the packages so that, for example, they all use the output path `dist`. For the backend, it is still `built` on `main` (I changed this in my PR, but the `.dockerignore` files are still duplicated and would need to be removed in the package / app pathes)

PR #53, opened 2024-10-08
chore(easier-setup): Easier start for new developers

JumpLink:
To make it easier for the new developer to get started, I have continued to work on the developer experience:

* Easy startup, where everything needed to start can be done in the root directory.
* The `docker-compose.yml` and the corresponding `.env` are now in the root directory.
* `vite` and `tsc` no longer clear the screen in watch mode, which previously caused important information to be lost in the console.
* To be able to disable the `clearScreen` option I also updated `vite` and the plugins used.
* README.md updated for an easy start.
* Added information to the backend`s README.md on how to make a test purchase with a test credit card.
* I have also added a few new scripts to the `package.json`, for example `presetup` which should be executed before the first start of the container stack and `setup` which should be executed when the container stack is running.


wpf500 (approved):

wpf500:
This is great. As a next step (in a separate PR), how about resolving #1 and moving the router into this repo? Then the docker-compose file can have it as a service and we can delete all the proxy config from the frontend.

PR #55, opened 2024-10-11
chore(github): Set deno to v1.x for Github Actions

JumpLink:
Deno v2 is released but we are using v1 so far. For v2 we need a separate PR.

This is a simper version of #54.


JumpLink:
Since this PR does not contain any changes but only enforces Deno v1 as we have used it so far, I am free to merge it directly

PR #56, opened 2024-10-11
chore(github): Fix action: Location of .env has changed

JumpLink:
Another GitHub Action error that only occurred in main and was therefore previously unnoticed similar to #55.

I have also moved the copying of the .env for the frontend to the package.json of the frontend, which should be clearer


JumpLink:
I have also merged this directly so that `main` will hopefully run without errors for future changes

PR #57, opened 2024-10-11
chore(router): Add router to monorepo

JumpLink:
# Integration of Router and Alignment of Docker Compose Stacks

## Overview
This PR integrates the router into the monorepo and aligns the [productive Docker Compose stack](https://github.com/beabee-communityrm/hive-deploy-stack) with the [developer Docker Compose stack](https://github.com/beabee-communityrm/monorepo/blob/main/docker-compose.yml). These changes aim to create a more consistent development environment that closely mirrors the production setup.

This solves the first bits of #1, next part would be to merge the two routers and nginx configurations into one, but before that, the old frontend should be removed.

### Router Integration
- Merged the router branch into this branch using `git subtree add --prefix=apps/router router main`
- Added the `app_router` service to the monorepo's Docker Compose configuration.
- This allows developers to test the frontend behavior through the router, simulating the production environment more accurately.

### Frontend Development
- Retained Vite serve + proxy for Hot Module Replacement (HMR) capabilities.
- The frontend can now be accessed via the `app_router` service, enabling testing of router-specific behaviors.

### API Access
- Modified the API access to be direct in the local environment, potentially improving debugging efficiency.
- Updated environment variables to reflect the new API access method.

### Documentation Updates
- Updated `README.md` with:
  - Explanations of the different open ports and their purposes.
  - New instructions for setting up and using the updated development environment.

### Environment Configuration
- Moved backend environment file from `./apps/backend/.env` to `./.env` in the root directory.
- Updated `.env.example` and `./apps/frontend/.env.example` to reflect new configuration requirements.

## Additional Changes
- Fixed Prettier VSCode configuration
- Disabled Deno for `packages/core`
- Further comments with the hope of making the different frontend configurations for Vite and Docker easier to understand

## Required Actions for Developers
After merging this PR, developers need to update their local environment:

1. Move `./apps/backend/.env` to `./.env` in the root directory.
2. Update `./.env` according to the changes in `.env.example`.
3. Modify `./apps/frontend/.env` based on the updates in `./apps/frontend/.env.example`.


JumpLink:
@wpf500 With the last two commits, I have now implemented what we discussed. The port to the legacy router is no longer open and used and the Vite proxy is back in place, but the environment variables are now a bit more clearly defined, I think.

Running HMR via Docker is theoretically possible, but all the approaches I've found don't convince me and are complicated in their own way. Therefore, I think that the proxy is the best solution for now.

docker-compose.yml
```diff
@@ -37,6 +39,9 @@ services:
   api_app:
     <<: *base-app
     command: node ./built/api/app.js
+    ports:
+      # Only used for local development
+      - 3004:3000
```
wpf500, line 44: Why does the API app need to be exposed directly?

apps/backend/src/core/sessions.ts
```diff
@@ -16,7 +16,8 @@ export function setTrackingCookie(memberId: string, res: Response) {
     ...config.cookie,
     httpOnly: true,
     maxAge: 365 * 24 * 60 * 60 * 1000,
-    sameSite: "none"
+    // If secure is false, sameSite none is not allowed
+    sameSite: config.cookie.secure ? "none" : "lax"
```
wpf500, line 20: What does this change relate to?

wpf500 (commented):
Nice to have the router in the monorepo :). Shame about having to keep the Vite proxy but incremental improvements are still good!

One thing: there seems to be some unrelated changes here, but maybe I'm not understanding something? What are all the Prettier related changes about?

wpf500:
Ah I've actually just read in your PR description that the Prettier changes are separate :+1:

apps/backend/src/core/sessions.ts
JumpLink: I needed that when I wanted to remove the Vite Proxy. It can now be removed again, I will do that.

docker-compose.yml
JumpLink: It was designed to communicate directly with the API instead of via the proxy, but it can probably get out again now. I'll check

JumpLink:
> Ah I've actually just read in your PR description that the Prettier changes are separate 👍

To be honest, I still know exactly what you mean. There are `prettier` related changes in the PR, but they relate to VSCode, which never worked for me, and with the changes they do now.

docker-compose.yml
JumpLink: is removed now

apps/backend/src/core/sessions.ts
JumpLink: is removed now, too

JumpLink:
@wpf500 I can confirm that this PR works on ‘dev’.

The docker router image was published over the GitHub Action and is publish on docker hub with the `dev` tag🫰🏿👍:
grafik

And also runs in Portainer:
grafik

And the frontend [is still available](https://dev.beabee.io/) 👍.

PR #63, opened 2024-10-29
Feat(contact-tagging): Add support tagable contacts with bulk action

JumpLink:
# Contact Tagging System Implementation

## Description
This PR implements a comprehensive contact tagging system by aligning it with the existing Callout Responses tag functionality. The changes include:

- Removal of legacy tag implementation from ContactProfile
- New tag system aligned with Callout Responses tags
- New ContactTagController for managing contact tags
- Batch update capabilities for contacts
- Enhanced filtering with pagination support
- Database migrations for tag storage

## Key Changes
- Added ContactTagController with CRUD operations for global tags
- Implemented batch update functionality for multiple contacts
- Created new DTOs and transformers for tag management
- Updated filtering system to support tag-based queries
- Added database migrations for tag storage
- Aligned tag management UI components

### Tag Management
- [ ] Create, edit and delete tags
- [ ] Assign tags to individual contacts
- [ ] Batch assign tags to multiple contacts
- [ ] Remove tags from contacts
- [ ] Test tag search functionality

### Filtering & Pagination
- [ ] Test tag-based filtering with pagination
- [ ] Verify combined filters (tags + other filters)
- [ ] Test filter persistence across pages
- [ ] Verify performance with large datasets

### Integration Tests
- [ ] Verify Callout Response tag filtering still works
- [ ] Test tag migration from old system
- [ ] Verify API responses for tag operations
- [ ] Test error handling for invalid operations

### UI Components
- [ ] Test TagEditorForm functionality
- [ ] Verify ToggleTagButton behavior
- [ ] Test tag selection in filters
- [ ] Verify tag display in contact lists

## Related Components
- ContactTagController
- TagTransformer
- Contact filtering system
- Tag migration scripts
- Tag management UI

## Translations Checklist
* [x] Updated translations in Google Sheet
* [ ] Merge new translations in Google Sheet to main sheet


JumpLink:
@wpf500 I initially based my implementation on  `CalloutTag` , but then deviated from it. In the `ContactTag` model, I now only store the tags and no reference to the contact (`CalloutTag` stores the `calloutId`). Instead, I have created a `contact_tag_assignments` table using `@ManyToMany("ContactTag", "contact")` and `@JoinTable({...})`, which is used for the assignments. For callout responses, tags are managed per callout; for contacts, they are managed globally. Maybe you would like to take a look at my unfinished implementation? Feel free to give me suggestions on what else could be combined.

apps/backend/src/api/controllers/ContactController.ts
```diff
@@ -145,6 +143,35 @@ export class ContactController {
     return res;
   }
 
+  @Authorized("admin")
+  @Get("/tags")
```
wpf500, line 147: I think it might be cleaner if these endpoints sit in a different controller under `/contact-tags` rather than `/contacts`. What do you think?

apps/backend/src/api/transformers/ContactTransformer.ts
```diff
@@ -88,6 +92,12 @@ class ContactTransformer extends BaseContactTransformer<
         qb.innerJoinAndSelect(`${fieldPrefix}profile`, "profile");
       }
 
+      if (query.with?.includes(GetContactWith.Tags)) {
```
wpf500, line 95: This is another one of these "only knowledge in my head" moments.

This join can return more than one row, so it can't be done in the main query. The pagination logic assumes one row per contact, which wouldn't be the case if a contact had multiple tags. This would mean `OFFSET` no longer works as intended.

This is what `loadContactRoles` in `modifyItems` is all about. It loads all the relevant contact roles after the initial query. You need to fetch the contact tags in the same way.

packages/core/src/models/Contact.ts
```diff
@@ -76,6 +80,20 @@ export class Contact {
 
   contributionInfo?: ContributionInfo;
 
+  @ManyToMany("ContactTag", "contact")
+  @JoinTable({
```
wpf500, line 84: I think it would be better to explicitly define this table as a model (like `CalloutResponseTag`)?

It's more explicit and also means we can save other data about the relationship. In particular, it would be good to have a `@CreateDateColumn() date: Date` column which stores when the tag was assigned (again, see `CalloutResponseTag` for reference)

wpf500 (commented):
A few small comments but in general this is really great :tada:. Well on it's way to being one of those magical PRs that adds a feature with a negative line count :smile:

apps/backend/src/api/transformers/ContactTransformer.ts
JumpLink: Did I now do it like in the CalloutResponseTransformer

packages/core/src/models/Contact.ts
JumpLink: implemented now

apps/backend/src/api/controllers/ContactController.ts
JumpLink: @wpf500 I've changed it now, but now it's no longer consistent with the callout tags, so maybe they should get their own controller too.

apps/backend/src/api/controllers/CalloutResponseController.ts
```diff
@@ -58,7 +58,7 @@ export class CalloutResponseController {
   async updateCalloutResponse(
     @CurrentAuth({ required: true }) auth: AuthInfo,
     @Params() { id }: UUIDParams,
-    @PartialBody() data: CreateCalloutResponseDto // Should be Partial<CreateCalloutResponseData>
+    @PartialBody() data: Partial<CreateCalloutResponseDto>
```
wpf500, line 61: I think the type here has to be `CreateCalloutResponseDto` because generics break the reflection that `class-transformer` uses to create the class, which means this just becomes a generic object which doesn't get validated.

See:
https://github.com/typestack/class-transformer?tab=readme-ov-file#working-with-generics
https://stackoverflow.com/questions/57797381/validation-does-not-work-with-partialdto-nestjs

Really we should do what they suggest in those links, which is to have a helper which creates a concrete "partial" class that can be reflected against.

apps/backend/src/api/controllers/ContactController.ts
```diff
@@ -154,37 +183,35 @@ export class ContactController {
...
-        throw new UnauthorizedError();
-      }
-
-      await ContactsService.updateContactProfile(target, data.profile);
-    }
-
-    return await ContactTransformer.fetchOneById(auth, target.id, {
-      with: data.profile ? [GetContactWith.Profile] : []
-    });
+    return await ContactTransformer.updateOneByContact(auth, target, data);
```
wpf500, line 214: Thanks for cleaning this up :tada:

apps/backend/src/api/controllers/ContactTagController.ts
```diff
@@ -0,0 +1,115 @@
...
+   * Returns: [{ id: "...", name: "Important", description: "..." }, ...]
+   */
+  @Authorized("admin")
+  @Get("/")
+  async getAllContactTags(
+    @CurrentAuth({ required: true }) auth: AuthInfo
+  ): Promise<GetContactTagDto[]> {
+    const result = await contactTagTransformer.fetch(auth, {
+      limit: -1,
+      rules: {
```
wpf500, line 49: I think you can just not declare `rules` here no?

```ts
{ limit: -1 }
```

wpf500 (commented):
This is looking amazing so far. Thanks for adding so much documentation, cleaning up the code, giving it a better structure... on top of implementing the tagging in such a way that we could easily add more taggable entities in the future.

apps/backend/src/api/controllers/CalloutResponseController.ts
JumpLink: Implemented in this commit: https://github.com/beabee-communityrm/monorepo/pull/63/commits/1925031e425b81d0e1e093938d730f0378a929eb

apps/backend/src/api/controllers/ContactTagController.ts
JumpLink: Yes, thank you

JumpLink:
> This is looking amazing so far. Thanks for adding so much documentation, cleaning up the code, giving it a better structure... on top of implementing the tagging in such a way that we could easily add more taggable entities in the future.

I was very happy to do so, I found it difficult to understand (and I still haven't fully understood it) but the tidying up and documenting helped me to do so, so it wasn't entirely self-serving ;)

apps/backend/src/api/controllers/CalloutResponseController.ts
JumpLink: I have now undone that and simply created a new DTO class instead.

apps/backend/src/api/dto/CalloutResponseDto.ts
```diff
@@ -144,6 +144,35 @@ export class CreateCalloutResponseDto {
   assigneeId?: string | null;
 }
 
+export class UpdateCalloutResponseDto
+  implements Partial<CreateCalloutResponseDto>
+{
+  // TODO: validate
```
wpf500, line 150: This is good for now, but at some point we should look into using some helpers like the NestJS ones to avoid all the boilerplate.

https://docs.nestjs.com/openapi/mapped-types

apps/backend/src/api/transformers/ContactExporter.ts
```diff
@@ -7,7 +7,7 @@ import { getMembershipStatus } from "@beabee/core/services/PaymentService";
 import { GetExportQuery } from "@api/dto/BaseDto";
 import { ExportContactDto } from "@api/dto/ContactDto";
 
-import { Contact } from "@beabee/core/models";
+import { Contact, ContactTagAssignment } from "@beabee/core/models";
```
wpf500, line 10: Unused import can be removed

apps/backend/src/tools/database/anonymisers/models.ts
```diff
@@ -298,8 +299,9 @@ export const contactProfileAnonymiser = createModelAnonymiser(ContactProfile, {
     line2: () => chance.pickone(["Cabot", "Easton", "Southmead", "Hanham"]),
     city: () => "Bristol",
     postcode: () => "BS1 1AA"
-  },
-  tags: (tags) => tags.map(() => chance.profession())
+  }
+  // TODO: Move to contactAnonymiser?
```
wpf500, line 303: Is this TODO still relevant?

apps/backend/src/tools/database/anonymisers/models.ts
```diff
@@ -391,3 +393,11 @@ export const segmentContactsAnonymiser = createModelAnonymiser(SegmentContact, {
 
 export const segmentOngoingEmailsAnonymiser =
   createModelAnonymiser(SegmentOngoingEmail);
+
+export const contactTagAssignmentAnonymiser = createModelAnonymiser(
+  ContactTagAssignment,
+  {
+    contactId: () => uuidv4(),
+    tagId: () => uuidv4()
```
wpf500, line 401: Can you move this up as this file is generally (hopefully) in alphabetical order.

We need an anonymiser for the contact tag table too, otherwise the export/import will break as the `tagId` in this table won't be a valid foreign key

Something like
```ts
const contactTagAnonymiser = createModelAnonymiser(ContactTag, {
  id: () => uuidv4(),
  name: () => chance.word(),
  description: () => chance.sentence()
});

File: apps/backend/src/api/transformers/BaseContactTransformer.ts [hunk 11]
```diff
@@ -217,3 +202,44 @@ const calloutsFilterHandler: FilterHandler = (qb, args) => {
 
   return { calloutId, ...params };
 };
+
+function tagField(): FilterHandler {
```
wpf500, line 206: I think this can be deleted now?

wpf500 (commented):
A few minor comments, but generally looks really great :tada:

wpf500:
One other question: have you tested the migration against an instance that had old style tags to migrate? Did it work on dev for example?

apps/backend/src/tools/database/anonymisers/models.ts
JumpLink: This comes from your last change where you anonymised the old tags, I wasn't sure if I should do something similar, hadn't looked at it more closely, but actually the tags should only be output if this is specified with "with"

JumpLink:
> One other question: have you tested the migration against an instance that had old style tags to migrate? Did it work on dev for example?

Yes you can look at this on dev, the old tags have been successfully preserved

apps/backend/src/api/transformers/BaseContactTransformer.ts
JumpLink: Yes, thank you

apps/backend/src/tools/database/anonymisers/models.ts
JumpLink: I added it and included it in the exports in `src/tools/database/export.ts`. Do I need to do anything else?

apps/backend/src/tools/database/anonymisers/models.ts
JumpLink: I have now removed the property tags here.

apps/backend/src/api/dto/CalloutResponseDto.ts
JumpLink: This helper can be found here but requires other helpers to: https://github.com/nestjs/mapped-types/blob/master/lib/partial-type.helper.ts

packages/core/src/errors/DuplicateTagNameError.ts
```diff
@@ -0,0 +1,17 @@
...
+export class DuplicateTagNameError extends BadRequestError {
+  readonly code = "duplicate-tag-name";
+
+  constructor(tagName: string) {
+    super(`Tag with name "${tagName}" already exists`);
+    Object.setPrototypeOf(this, DuplicateTagNameError.prototype);
+  }
+
+  static isPostgresError(error: any): boolean {
+    return (
```
wpf500, line 12: You should use `isDuplicateError` here with the field name rather than the constraint ID.

In theory the constraint ID should be a cleaner way of checking this, but at the moment it's autogenerated by TypeORM and there's no guarantee it will stay the same.

wpf500 (changes requested):

JumpLink:
With my lastest changes I have implemented:
* Ordered tags
* Wrap/break tags for items with many tags
* Improved delete tag translation
* Better error message on duplicate tag names
* Readd callout filter group for contacts
* Do not disable the toggle button if no item is selected to be able to reach the tag manager without a selection
* Keep the selection to be able to toggle multiple tags for the same selected items
* Fixed Tag Manager route from detail pages
* Make tags clickable

PR #64, opened 2024-10-30
Fix various Docker stack problems

wpf500:
To merge #54 we need to change the `docker-compose.yml` in https://github.com/beabee-communityrm/beabee-deploy/ and https://github.com/beabee-communityrm/hive-deploy-stack.

Given that we have to make changes anyway, this is an opportunity to fix various problems we have in the Docker stack configuration. In general we have too much configuration in the `docker-compose.yml` which should be in the Docker images themselves.

**Problem:** Legacy app, API, webhook and cron services use the same image, but override the command
**Fix:** Each service should have it's own image, the command should be in the build step

**Problem:** Two routers, a legacy one and a new router
**Fix:** Merge these into one and delete the legacy router service. Note: legacy /static content is now served by the legacy app, this is preferable to transferring this route to the main app, and in any case is on a path to deletion.

**Problem:** `init` flag is required configuration
**Fix:** tini should be built into the images ([read more](https://github.com/krallin/tini?tab=readme-ov-file))

This PR renames the our Docker images too. This is partly for consistency, but also to be cautious around accidentally breaking something for self hosting users. They will need to update their Docker stack configuration to opt into these changes
* `beabee/beabee` split into:
  * `beabee/beabee-legacy-app`
  * `beabee/beabee-api-app`
  * `beabee/beabee-webhook-app`
  * `beabee/beabee-cron-app`
* `beabee/app-router` renamed to `beabee/beabee-router`
* `beabee/frontend` renamed to `beabee/beabee-frontend`
* `beabee/router` removed (legacy router)

Corresponding changes for the stack configuration are here:
* https://github.com/beabee-communityrm/hive-deploy-stack/pull/6
* https://github.com/beabee-communityrm/beabee-deploy/pull/17

**Note:** https://github.com/beabee-communityrm/monorepo/pull/65 must be merged into this branch first, as it fixes the release deployment


JumpLink:
Looking at the Dockerfile, we can simplify those multiple build steps into something cleaner:

Currently:
```dockerfile
# Build common
COPY packages/common /opt/packages/common
RUN yarn workspace @beabee/beabee-common build

# Build core
COPY packages/core /opt/packages/core
RUN yarn workspace @beabee/core build

# Build backend
COPY apps/backend /opt/apps/backend
RUN yarn workspace @beabee/backend build
```

We can replace this with:
```dockerfile
# Copy all workspace packages needed for backend
COPY packages /opt/packages

# And the app we want to build
COPY apps/backend /opt/apps/backend

# Build backend and its dependencies in one go
RUN yarn workspaces foreach --recursive --topological-dev --parallel --verbose --from @beabee/backend run build
```

The command:
- `--recursive`: Goes through all dependencies recursively
- `--topological-dev`: Makes sure everything builds in the right order based on dependencies
- `--parallel`: Builds stuff in parallel for speed (we can remove this if it causes issues)
- `--verbose`: Shows detailed output so we can debug if needed
- `--from @beabee/backend`: Starts from backend and builds everything it needs

This way we only need one build command and it'll handle the dependencies automatically. Plus if we add new packages later without adapt the docker image for that new packages - the build command will pick them up automatically.

We can also do the same for the frontend. This comment also applies to #54 if this PR is just an experiment

_In my PR, which you rejected, I went one step further and created my own base image for building the complete workspace, on which the backend and frontend were then set up. This had the advantage that all packages only had to be built once. We currently build common dependencies such as `beabee-common` once for the backend and once for the frontend. But maybe there are also disadvantages in the image size._

JumpLink (approved):
The PR is great and I can't see any flaws. My LLM seems to agree:

> This PR represents a significant improvement to the Docker infrastructure. The changes are well-organized, properly scoped, and show careful consideration for both technical excellence and user impact.

The only thing would be a migration documentation, which we could perhaps write in the release description.

wpf500:
Merging this into #54

PR #65, opened 2024-11-01
Merge all deploy/release actions

wpf500:
Merge deploy and release actions into one flow.

This has the following benefits:
- All Docker images use exactly the same tagging schema
- Remove code duplication from deploy and release steps
- Use only one version file in hive-deploy-stack, rather than separate ones for the frontend, API and router. These projects are now all in the same monorepo so their version numbers are the same.
- Resolves a race condition where the backend or frontend release could update hive-deploy-stack at the same time and cause the other to fail


JumpLink (approved):
Thanks, much better! 

In the project root, you can also simply run `yarn check`, which then calls all the checks for the packages and apps and processes them in parallel. Maybe not so well suited as a Github Action because you might prefer to have these things listed individually, but it would have the advantage of automatically including new packages. Just a suggestion, but I'm happy with the PR

wpf500:
Ah I just copied that from the old check actions, I am more than happy to unify that into yarn check :).

wpf500:
Merging this into `fix/docker-images` now

PR #66, opened 2024-11-04
fix: callout response anonymisers

wpf500:
This PR adds some documentation to the anonymisation code and fixes the callout response anonymisers.

How to run anonymisers?

You can run the `tools/database/export` tool on any instance. Typically you would do this by:
```bash
docker compose run -T --rm app node built/tools/database/export > export.data
```
There is a bug for instances with dev mode enabled (`BEABEE_DEV=true`). Dev mode enables query logging, but TypeORM improperly logs to stdout instead of stderr. This means the logging is saved in `export.data` which corrupts the file format.

The simplest way around this for now is to exclude those lines from the file
```bash
docker compose run -T --rm app node built/tools/database/export | grep -v ^query: > export.data
```

`tools/database/export-demo` is a smaller export which only exports a subset of tables, more suitable for updating a demo instance.

Once you have an export, you can import it locally using the following command
```bash
docker compose run -T --rm app node built/tools/database/import < export.data
```

Be warned this will clear all the current data on your database. You must import an export that has been generated by an instance with a compatible database schema.


wpf500:
@JumpLink I'm merging this as I need to use it for my Lokal map import

PR #69, opened 2024-11-05
Update deploy.yml: fix/actions -> main

JumpLink:
Maybe it was an accident? I just noticed it.


wpf500:
I'm still testing the actions at the moment, had a problem with pushing multiple tags (which only happens on a release)

PR #75, opened 2024-11-11
fix: cron job service

wpf500:
This PR fixes the cron job service, which had the following problems
1. No env vars passed to the cron jobs
2. `node` couldn't be found in cronjobs (needs modified `PATH`)
3. Logs were going to the default cron log location (/var/mail) rather than stdout

To make cron jobs work properly we need to
1. Store the environment variables in `/etc/environment`
2. Add `/usr/local/bin` to the `PATH`
3. Install rsyslog and redirect cron logs to `/proc/1/fd/1`. This is stdout for the main process (PID 1).

### DIFF
--- a/apps/backend/Dockerfile
+++ b/apps/backend/Dockerfile
@@ -96,10 +96,18 @@ CMD [ "node", "./built/webhooks/app.js" ]
 ## Cron image
 FROM dist AS cron_app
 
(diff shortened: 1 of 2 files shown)


wpf500:
@JumpLink I'm merging this now to fix the jobs, but I'd appreciate a post-deploy review

PR #76, opened 2024-11-12
feat: import response tool

wpf500:
This tool allows you to batch upload callout responses to a callout from a CSV

```bash
node import-callout-responses <callout-slug> < <csvfile>
```


JumpLink:
I would prefer it if we started a CLI package (I had also done that in my big PR) and combined all the scripts in it so that we have a CLI that can do different things (database migration, user creation, help. etc). In this regard, I have had very good experiences with [yargs](https://yargs.js.org/), e.g. in [this private project](https://github.com/gjsify/ts-for-gir/tree/main/packages/cli).

PR #82, opened 2024-11-18
refactor: share backend config

wpf500:
Store the shared backend build configuration in a common location. For now this is just the Prettier config and TS build options. I also used the opportunity to simplify some duplication in the `.gitignore` files.


JumpLink (dismissed):
@wpf500 Great, we could also make a package out of it by adding a package.json and exporting the config files in it and adding the whole thing to the workspace. This would be comparable to other packages like [@tsconfig/node22](https://www.npmjs.com/package/@tsconfig/node22). Another advantage is that the configs can then also have dependencies, for example prettier plugins. It probably also makes sense for each config to be its own package.

I would then move it to the packages folder. Later, we could also publish such packages on npm, so that new projects (e.g. projects that are not yet part of the monorepos because they are experimental, for example) could still use the same configs.

JumpLink:
@wpf500 I have now created two new packages, ‘@beabee/prettier-config’ and ‘@beabee/tsconfig’. Both deliver their dependencies themselves, which means that they no longer have to be installed separately in the packages.
In addition, the core package has not yet been checked with the check `script`, but I have changed that.

In addition, I updated `@tsconfig/node20` to `@tsconfig/node22`, which revealed new type errors that I was able to fix by updating to `@tiptap` to `v2.10.0`. I tested the RichTextEditor and it still works as expected.

Currently, I have an error building the Dockerfiles, so I'll take a look at that now.

PR #83, opened 2024-11-18
refactor: move webhooks to separate app

wpf500:
This PR moves the webhook app into it's own app, part of the ongoing work to split the backend app up.

Most changes relate to updating the import in the tools as the `runApp` helper has been moved to `@beabee/core`. These tools will be moved to their own app soon.

The only other significant change is to the `Dockerfile`. I've just moved the file to a common location and modified it to build different parts of the backend. A cleaner way would be to unify backend and frontend Docker builds with a common base image (as in #4), but this way is a clean way to share the backend build configuration without more major changes.

Note: #82 should be merged first


JumpLink (commented):
I think it's a great approach, but I also think that the Dockerfile should be its own NPM package. 

I think in the future this repo should also be part of the workspace if we found a clean way for that: https://github.com/beabee-communityrm/hive-deploy-stack and this repo contains not only the `docker-compose.yml` but also scripts, so it makes sense if it is its own package that can be executed together during the build process and for the Dockerfile it should also be so that it is consistent.

PR #86, opened 2024-11-19
Hotfix(feat/contact-tagging): Delete tag rules with with unused tags

JumpLink:
This removes tag segment rules with unused (and not existing) tags

### DIFF
--- a/packages/core/src/migrations/1731670066414-MigrateSegmentTagRules.ts
+++ b/packages/core/src/migrations/1731670066414-MigrateSegmentTagRules.ts
@@ -75,7 +75,7 @@ export class MigrateSegmentTagRules1731670066414 implements MigrationInterface {
                 }
               }
             } else {
-              processRules(rule.rules);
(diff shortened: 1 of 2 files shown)


packages/core/src/migrations/1731670066414-MigrateSegmentTagRules.ts
```diff
@@ -52,17 +69,20 @@ export class MigrateSegmentTagRules1731670066414 implements MigrationInterface {
...
                       `Tag not found: ${tagName} in segment: ${segment.id}`
                     );
-                    return value;
+                    modified = true;
+                    return "__DELETE_ME__";
                   });
                 }
               }
             } else {
               processRules(rule.rules);
```
wpf500, line 78: I guess this needs to catch the result?
```ts
rule.rules = processRules(rule.rules);
```

JumpLink: oh yes, thank you

PR #89, opened 2024-11-20
feat: add callout reviewer

wpf500:
This PR introduces callout reviewers, allowing users to be given the necessary permissions to see callout responses for one or more specific callouts.

They can also use most of the associated response features: tagging, commenting, bucketing, editing, etc.

### Changes

`BaseTransformer` now has base methods for creating, updating and deleting entities. This makes it very easy to implement a CRUD interface for any entity, while fully supporting rule group based permissions and filtering logic.

`TagTransformer` and descendents have been refactored to use abstract and inherited classes respectively. This allows tag transformers to override transformer methods like any other transformer, which was necessary to introduce different permissions for contact and callout tags. In the process the PR also cleans up some of the external references to the assignment model and IDs, this can all be managed internally.

Authentication was previously represented by `AuthInfo | undefined`. This has been refactored by introducing `AuthInfoNone` to represent undefined, eliminating the ambiguity between `undefined` and optional parameters. This was particularly important as inherited transformers often only implement a subset of undefined-able method parameters. With this implementation the programmer is force to provide an `AuthInfo`, even if it's just `AuthInfoNone`.

### Next steps

Removing the remaining direct database access calls from the controllers (e.g. with `getRepository`), this should all be done using the `BaseTransformer` CRUD methods.

`CurrentAuth` can eventually completely replace the simpler `Authorized` controller decorator. All permission logic would move to the transformers, and where appropriate into the data layer.

Refactoring the `ContactTransformer` and `CalloutResponseTransformer` classes to use the standard `update` methods.

Refactoring the query building code to remove `fieldPrefix`. `fieldPrefix` was introduced to fix a limitation in using table aliases in `UPDATE` queries, but the new `update` method uses a much cleaner implementation that doesn't have this limitation. It was always a horrible hack and is very confusing.

### To do
- [x] Make UI more friendly for reviewers
- [x] Add UI for adding reviewers
- [x] Allow reviewers to see and add comments
- [x] Allow reviewers to see and assign reviewers


JumpLink (approved):
Very nice :)

PR #99, opened 2024-12-03
Remove Deno

JumpLink:
Deno is great and I would still prefer it over Node.js for new projects. But mixing the two environments just leads to too many problems and frustrations. So this PR removes the Deno dependency and switches the `common` and `client` packages to Node.js.


wpf500:
I don't really understand why all the imports have been changed from `.ts` to `.js` but I'm sure there's a good explanation :upside_down_face:

JumpLink:
> I don't really understand why all the imports have been changed from `.ts` to `.js` but I'm sure there's a good explanation 🙃

Deno needs the `.ts` extension if you want to work directly with typescript, but `tsc` only allows this with the `allowTsExtension` option (or something like that), which might work thanks to esbuild (which can handle this), but might cause problems if it is inconsistent and you want to list all references in VSCode. This is just a guess, I haven't tested it, but that's why I decided to change it to `.js'.

PR #106, opened 2024-12-11
fix: try to handle Mailchimp compliance state

wpf500:
If a user is unsubscribed in Mailchimp then this will now put them in "pending" mode. Mailchimp sends them a confirmation email to opt-in, then the webhook will update them to subscribed when they do this.


JumpLink (approved):
Thanks for the improvements! I've reviewed the changes and they look good. The error handling for Mailchimp compliance states and the overall refactoring are solid improvements.

I've already resolved two TODOs:
- ✅ `TODO: this should be done in the newsletter service`
- ✅ `TODO: Update newsletter status in database`

There's still one TODO remaining that should be addressed in a follow-up PR:
```typescript
// TODO: should be an update without status
//       currently we're sending the status unnecessarily when only updating fields
```

**Approving with Follow-up:**
The current changes can be merged, but I will create a new ticket #115 to handle the remaining TODO.

PR #114, opened 2024-12-20
build: update dependencies

wpf500:
Done
* Upgraded ESLint to v9, migrated to `eslint.config.mjs`, and fixed ESLint errors
* Upgraded most minor/patch dependencies
* Upgraded some major versions

To do
* `@intlify/unplugin-vue-i18n` and `vue-i18n`
* `@maptiler/geocoding-control`
* `maplibre-gl`
* `node-fetch`

Holding back
* `@stripe/stripe-js` and `stripe`: possibly includes API version upgrades, need more time to look at this
* `@types/express`: upgrade from Express v4 to v5, [currently tagged](https://expressjs.com/2024/10/15/v5-release.html) as `next` rather than `latest`

Ignore as only in legacy app:
* `query-string`: latest version is an ESM, not compatible with legacy app
* `gulp`, `gulp-sass`, `gulp-postcss`
* `ajv-formats`
* `helmet`
* `sass`: includes changes to SASS processing I think, not worth it


JumpLink (approved):
Thanks, I have tested the chages and they are working.

This packages we should add to the update TODO list, too:
* `csurf` - because this package is marked as deprecated
* `vite-plugin-pages` - Because this package is replaced by [unplugin-vue-router](https://github.com/posva/unplugin-vue-router)

PR #119, opened 2025-01-06
feat: initial implementation of showing clustered responses

wpf500:
This PR adds the ability to see clusters of responses on the Atlas. Previously if multiple responses were too close together you could only see one of them.

This feature respects the Atlas max zoom setting which allows admins to control the maximum "declustering" zoom point. Previously it was hard coded that all clusters disappeared at a certain zoom level because you could only see responses on their own, not in a cluster.

grafik


apps/frontend/src/pages/callouts/[id]/map.vue
```diff
@@ -420,55 +478,105 @@ async function handleAddClick(e: { event: MapMouseEvent; map: Map }) {
...
+ * @param cluster The cluster
+ * @param map The map object
+ */
+function handleClusterClick(cluster: MapClusterFeature) {
+  const mapSchema = props.callout.responseViewSchema?.map;
+  if (!map.value || !mapSchema) return; // Can't actually happen
+
+  const source = map.value.getSource('responses') as GeoJSONSource;
+  source.getClusterExpansionZoom(cluster.properties.cluster_id, (err, zoom) => {
+    if (err || zoom == null) return;
```
Comment (JumpLink) [inline apps/frontend/src/pages/callouts/[id]/map.vue:493]: Maybe adding error logging here like this?

```ts
if (err || zoom == null) {
  console.error('Failed to get cluster expansion zoom:', err);
  return;
}
```

apps/frontend/src/pages/callouts/[id]/map.vue
```diff
@@ -285,112 +294,161 @@ const newResponseAddress = computed(() => {
...
+/**
+ * Find the cluster or point that corresponds to the selected response number
+ *
+ * @param responseNo The response number
+ */
+function findSelectedFeature(responseNo: number) {
+  if (!map.value) return;
+
+  const feature = map.value.queryRenderedFeatures({
+    layers: ['unclustered-points', 'clusters'],
```
Comment (JumpLink) [inline apps/frontend/src/pages/callouts/[id]/map.vue:349]: Move magic strings to constants:

```ts
const LAYER_IDS = {
  CLUSTERS: 'clusters',
  UNCLUSTERED_POINTS: 'unclustered-points',
  SELECTED_RESPONSE: 'selected-response'
} as const;
```

apps/frontend/src/pages/callouts/[id]/map.vue
```diff
@@ -285,112 +294,161 @@ const newResponseAddress = computed(() => {
...
+
+  const responseNumbers = selectedFeatureResponses
+    // Remove first < and last >
+    .substring(1, selectedFeatureResponses.length - 1)
+    // Split each response number
+    .split('><')
+    .map(Number)
+    .sort();
+
+  return responses.value.filter((r) => responseNumbers.includes(r.number));
```
Comment (JumpLink) [inline apps/frontend/src/pages/callouts/[id]/map.vue:337]: My coding assistant suggests a performance optimization here. Your current implementation has a complexity of O(n) lookup, but you can use Map for O(1) lookups:

```ts
const responseMap = new Map(responses.value.map(r => [r.number, r]));
const selectedResponses = responseNumbers.map(n => responseMap.get(n)).filter(Boolean);
```

JumpLink (approved):
Great work! The code shows significant improvements in both functionality and structure 🚀 

The code is ready to merge as is, but feel free to implement my suggestions before or after merge.

apps/frontend/src/pages/callouts/[id]/map.vue
Comment (wpf500) [reply apps/frontend/src/pages/callouts/[id]/map.vue]: I've used a slightly different method to suggested, but the same principle

JumpLink (approved):
LGTM thank you!

PR #120, opened 2025-01-07
fix: prettier formatting in VSCode

wpf500:
I don't really know why, it might be related to https://github.com/prettier/prettier-vscode/issues/3066, but my Prettier stopped working in VSCode . It was working fine on the command line, e.g. `yarn prettier -c .`. The console in VSCode just reported "Invalid configuration".

A frustrating amount of debugging later, the change in this PR seems to fix it :shrug:.


JumpLink (approved):

JumpLink:
@wpf500 One additional note that only now comes to mind. I think it is more robust to import the plugins and pass them directly instead of passing the name

PR #124, opened 2025-01-13
feat: further Mailchimp integration improvements

### LINKED ISSUE #115: Update newsletter field updates to avoid sending status
Currently when updating contact fields in the newsletter provider, we're unnecessarily sending the status along with the field updates. This could potentially cause unintended status changes.

@wpf500 This is the last remaining TODO from PR [#106](https://github.com/beabee-communityrm/monorepo/pull/106).

**Current behavior**
In updateContactFields(), we're sending the status even when only updating fields:
~~~typescript
await this.provider.upsertContact({
  email: nlUpdate.email,
  status: nlUpdate.status, // Unnecessary status update
  fields
});
~~~

**Proposed solution**
Add a dedicated method for field-only updates in the `NewsletterProvider` interface:
~~~typescript
interface NewsletterProvider {
  updateContactFields(email: string, fields: Record<string, string>): Promise<void>;
}
~~~

**Tasks**
- [x] Add new method to `NewsletterProvider` interface
- [x] Implement method in `MailchimpProvider` and `NoneProvider`
- [x] Update `NewsletterService.updateContactFields` to use new method

**Priority**
Low - This is a cleanup task that doesn't affect functionality but would make the code more maintainable.

**Related**
- PR [#106](https://github.com/beabee-communityrm/monorepo/pull/106) - fix: try to handle Mailchimp compliance state

wpf500:
The `NewsletterService` and its providers now handle the newsletter status, groups and active membership status internally, where previously this logic was scattered around the codebase (in `ContactService` and others). There are still improvements to be made here, but this is a good step towards clearer separation of concerns.

Other changes:
* Documentation for lots of methods
* Resolves #115 by adding a new method `updateContactFields` which can update a contact's merge fields without needing to know the correct subscription status.
* Refactoring types into `types/`, as is the general trend across the codebase
* Refactoring underlying Mailchimp functionality into `lib/mailchimp` for better separation of concerns

Future improvements could be:
* Create a separate `NewsletterContact` entity which removes the `NewsletterService` dependency on `ContactProfile` for newsletter status syncing
* Using an event system to trigger newsletter updates, rather than hooking manually into `ContactsService`
* Separate 
* Moving the newsletter resync tool out of the legacy app


wpf500:
https://github.com/beabee-communityrm/monorepo/pull/124/commits/192bcd5d1033a067d7d14b8db6635857cbf5bc69 solves #115

PR #125, opened 2025-01-16
Switch from Jest to Vitest

JumpLink:
I switched to vitest for the client and the (future) e2e tests to try it out. The tests also use - as in #118 - `testcontainers` to start our Docker Compose test stack beforehand and it works very similarly to the jest configuration.

I came to the following conclusion:
* The tests actually run significantly faster (starting the docker compose stack is the biggest bottleneck here)
* The configuration is much simpler and clearer
* TypeScript is supported out-of-the-box
* `vitest` is the only dependency we need (no `ts-jest` or similar)
* Conversion was straightforward 

So I would suggest that I also replace the rest of the tests 👍


JumpLink:
@wpf500 I have now replaced all the jest tests with vitest

wpf500:
Woop :tada:

PR #128, opened 2025-01-17
Add client tests

JumpLink:
This PR introduces initial client-side tests to ensure proper functionality of our API client methods before implementing end-to-end API scenario tests. 

# Added new auth/info endpoint for authentication method verification

I added a new `auth/info` endpoint to provide a way to determine the current authentication status in our tests. This endpoint allows us to verify whether a client is authenticated via:
- API key
- User credentials
- Not authenticated

# Fixed errors and improved client implementation

I have fixed several issues in the client code and added comprehensive tests:
- Data and parameter handling in fetch
- Cache polyfill for fetch on Node.js to fix get the client login method working
- Better url handling

- Added tests to verify correct behavior of:
  - Contact creation and management
  - Role assignment and updates
  - Data serialization and deserialization
  - Edge cases like null/undefined handling

The tests now ensure that the client behaves correctly and consistently, especially for complex operations like role management.


packages/client/test/api/contact-role.test.ts
```diff
@@ -0,0 +1,159 @@
...
+      const updateData: UpdateContactRoleData = {
+        dateAdded: new Date(),
+        dateExpires: "" as any
+      };
+
+      await expect(
+        contactClient.role.update(testContactId, role, updateData)
+      ).rejects.toThrow();
+    });
+  });
```
JumpLink, line 111: @wpf500 Can you take a look at this?

wpf500: Hmm interesting, yeah I think the behaviour you are expecting is correct. There is a bit of a long standing issue with the way the API handles `undefined` and `null`, which we could chat about. But in any case JSON doesn't have an `undefined` equivalent, so `axios` should just strip it from the request body, so that's not the problem here.

The empty string also isn't throwing an error then?

JumpLink: An empty string leads to an error (see test below this one) I have only described it for completeness with how it should behave

JumpLink: I think the problem is more how `null` is handled. I think `null` should mean that the existing expiry date should be removed. `undefined` can remain as they are: The existing expiry date will not be changed.

wpf500: Yup that's a bug that should be fixed :+1:

apps/backend/src/api/transformers/AuthTransformer.ts
```diff
@@ -0,0 +1,36 @@
...
+import { Contact } from "@beabee/core/models";
+import { GetAuthInfoDto } from "@api/dto";
+
+class AuthTransformer {
+  /**
+   * Converts auth info and transforms nested contact data
+   * @param auth - The raw auth info
+   * @returns Transformed auth info with properly converted contact data
+   */
+  convert(auth: AuthInfo): GetAuthInfoDto {
```
wpf500, line 12: I think this needs a decorator to turn it into a `GetAuthInfoDto` no?

```
  @TransformPlainToInstance(GetAuthInfoDto)
```

packages/client/test/api/contact-role.test.ts
JumpLink: Should be fixed with my latest commit

packages/client/src/utils/cookie-polyfill.ts
wpf500, line 1: What is this for? I don't think we should support using cookies in Node environments. In general the aim is to get rid of using cookies for API authentication completely, it's only like this in the frontend at the moment because the cookie is shared between the legacy app and the new frontend.

apps/frontend/src/components/role/RoleEditor.vue
```diff
@@ -163,9 +164,9 @@ function roleToFormData(
     role: role?.role || ('' as const),
     ...(role
       ? {
-          startDate: format(role.dateAdded, 'yyyy-MM-dd'),
-          startTime: format(role.dateAdded, 'HH:mm'),
-          hasStartDate: true,
+          startDate: role.dateAdded ? format(role.dateAdded, 'yyyy-MM-dd') : '',
```
wpf500, line 167: Not sure I understand these changes, doesn't a role always have to have a `dateAdded`?

wpf500 (commented):
A few comments, but generally a great step towards actually having tests :tada:

packages/client/src/utils/cookie-polyfill.ts
JumpLink: I needed this to be able to log in as a user in the tests; I could just use the polyfill in the test instead, but I think it would be good if third-party developers could also develop applications for node.js with the client? Or do you think it's more of a problem if the client can also be used to authenticate as a user? axios, which is also used on the frontend, has support for cookies.

apps/frontend/src/components/role/RoleEditor.vue
JumpLink: The types used here are based on the update type, and for updates the properties are optional, so an error was thrown here, so the type should actually be fixed

apps/backend/src/api/transformers/AuthTransformer.ts
JumpLink: It worked, but I think you're right anyway.

packages/client/src/utils/cookie-polyfill.ts
wpf500: I guess it's fine for now, and once we've removed cookie based authentication from the API we can clean this up now :+1:

apps/frontend/src/components/role/RoleEditor.vue
wpf500: I have reverted the changes to this file now the types are fixed

wpf500 (approved):

JumpLink:
@wpf500 I have now added the STRIPE `.env` variables to the GitHub Repo Secrets and adjusted the workflow so that it sets them in the `.env`. I'm using a temporary PR #144 to test whether the test workflow will now run

JumpLink:
I have now tested the new client and did the following:
* Created contact tags
* Filtered by these tags
* Created a segment for the tag filter
* Changed various settings in the backend
* Uploaded a new logo and applied it 
* Created a new callout 
* I also edited the callout and added content with an image but received the following error from the backend: 

```
request entity too large {"expected":184503,"length":184503,"limit":102400, ...
```

I think the limit is a bit small because the image wasn't particularly large either, with a smaller image it then worked, but that's not a bug of the new client, so I'm ignoring that for now in this PR.

JumpLink:
@wpf500 I also tested the creation of a new member and fixed another bug in this regard by removing customerId from `completePaymentFlow` again, then that worked again too. So now most of the new client should be tested and it seems to work very well :)

wpf500:
> I also edited the callout and added content with an image but received the following error from the backend:
> 
> ```
> request entity too large {"expected":184503,"length":184503,"limit":102400, ...
> ```
> 
> I think the limit is a bit small because the image wasn't particularly large either, with a smaller image it then worked, but that's not a bug of the new client, so I'm ignoring that for now in this PR.

This is a known issue as images in the callout content are embedded inline rather than uploaded separately, safe to ignore for now

wpf500 (approved):

PR #132, opened 2025-01-27
feat: add support for iDEAL payments

wpf500:
Support the iDEAL payment method. This method is slightly more complicated than others because you need to convert it to a SEPA direct debit during the setup process.

More details here: https://docs.stripe.com/payments/ideal/set-up-payment

### DIFF
--- a/packages/core/src/lib/stripe.ts
+++ b/packages/core/src/lib/stripe.ts
@@ -295,10 +295,10 @@ export function paymentMethodToStripeType(
       return "bacs_debit";
     case PaymentMethod.StripePayPal:
       return "paypal";
+    case PaymentMethod.StripeIdeal:
+      return "ideal";
     case PaymentMethod.GoCardlessDirectDebit:
       return "bacs_debit";
-    default:
-      throw new Error("Unexpected payment method");
   }
 }
 
@@ -321,6 +321,8 @@ export function stripeTypeToPaymentMethod(
       return PaymentMethod.StripeBACS;
     case "paypal":
       return PaymentMethod.StripePayPal;
+    case "ideal":
+      return PaymentMethod.StripeIdeal;
     default:
       throw new Error("Unexpected Stripe payment type");
   }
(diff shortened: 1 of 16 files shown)


packages/core/src/lib/stripe.ts
```diff
@@ -295,10 +295,10 @@ export function paymentMethodToStripeType(
       return "bacs_debit";
     case PaymentMethod.StripePayPal:
       return "paypal";
+    case PaymentMethod.StripeIdeal:
+      return "ideal";
     case PaymentMethod.GoCardlessDirectDebit:
       return "bacs_debit";
-    default:
```
JumpLink, line 300: Why removing this?

packages/core/src/providers/payment-flow/StripeProvider.ts
```diff
@@ -32,16 +34,38 @@ class StripeProvider implements PaymentFlowProvider {
 
   async completePaymentFlow(joinFlow: JoinFlow): Promise<CompletedPaymentFlow> {
     const setupIntent = await stripe.setupIntents.retrieve(
```
JumpLink, line 36: Is it possible to extend the type for the latest intent here?

packages/core/src/providers/payment-flow/StripeProvider.ts
```diff
@@ -32,16 +34,38 @@ class StripeProvider implements PaymentFlowProvider {
...
 
     log.info("Fetched setup intent " + setupIntent.id);
 
-    const siPaymentMethodId = setupIntent.payment_method as string;
+    // iDEAL is a one time payment method, use setup intent to retrieve the SEPA
+    // debit payment method instead
+    // https://docs.stripe.com/payments/ideal/set-up-payment
+    if (paymentMethod === PaymentMethod.StripeIdeal) {
+      const latestAttempt =
+        setupIntent.latest_attempt as Stripe.SetupAttempt | null;
```
JumpLink, line 50: Define the type earlier if this is cleanly possible, see previous comment

JumpLink (approved):
LGTM just a few comments on the type definitions

packages/core/src/lib/stripe.ts
wpf500: This should be an exhaustive switch statement (i.e. handles all payment method types), so removing the default clause means the type checker will complain when a new payment method is added

packages/core/src/providers/payment-flow/StripeProvider.ts
wpf500: I don't think this is possible unfortunately. This is the way [Stripe recommends](https://github.com/stripe/stripe-node?tab=readme-ov-file#using-expand-with-typescript) doing it in their docs.

PR #133, opened 2025-01-27
build: use built in tsc and node watch support

wpf500:
I started this PR because the auto-reloading wasn't working across all workspaces.

This PR removes `tsc-watch` in favour of the built in `tsc` watch feature and replaces restarting containers by the built in `node` watch feature.

`tsc-watch` was only being used for it's `onSuccess` functionality to restart Docker containers after a build. But actually you can just restart the process, rather than the container itself, and since v22 `node` has that built in watch mode which does just this.

This reduces dependencies, always good, and also means there are fewer hardcoded references to `docker compose` commands. As a bonus the process restarting should also be cleverer as `node` only restarts when dependent files are changed.

One outstanding problem is we still don't have a dev mode for `@beabee/beabee-common`. It uses a build script, I don't really understand how to add the esbuild watch mode within it :)


JumpLink (approved):
Great, much better!
Adding the watch mode for esbuild [is easy](https://esbuild.github.io/api/#watch), I'll be happy to do that. Do we want to make a separate PR for that? Do you want to merge this here first or should the PR point to this branch?

wpf500:
Ah yeah I saw the setting but wasn't sure how that would interact with the custom `esbuild.ts` script we use. I'll merge this PR and we can add the ESBuild stuff to one of the upcoming sprints

PR #135, opened 2025-01-30
feat(backend-cli): Enhance CLI with new commands and local execution support

JumpLink:
This PR enhances the `backend-cli` tool with new commands for payment test data generation and API key management, while also adding support for local execution outside of Docker containers.

Note: Some of this features are back ported from #128 for better separation of the pull requests.

## Features
- New CLI Commands
  - Payment test data generation (WIP)
    - Configurable number of test payments
    - Configurable payment amounts and periods
    - Optional processing fee and proration
    - Ability to use existing contacts by email
  - API key management
    - List all API keys with detailed information
    - Shows creator, expiry, and creation dates
    - Create API keys for specific users
  - System configuration
    - Configure support email based on domain
    - Update payment methods in join content
  - Sync commands
    - Mailchimp sync with dry-run support
    - Segment sync template (requires further preparatory work that should not be part of this PR)
  - Process commands
    - Gift processing functionality
- Local CLI execution support
  - Automatic detection of Docker environment
  - Environment-specific database connections
  - Separate configurations for test and development
- Refactored provider structure in core package
  - Moved providers to dedicated directories
  - Updated exports to use named exports instead of default exports
  - Added deprecation notices for default exports
  - Improved type definitions and interfaces
- Enhanced webhook handling
  - Added `StripeWebhookEventHandler` class
  - Added `GoCardlessWebhookEventHandler` class
  - Renamed webhook apps for consistency
  - Improved event handling structure
- Added comprehensive payment system documentation
  - Architecture overview
  - Payment flows
  - Provider integrations
  - Webhook handling
  - Development setup guide
- Docker and environment improvements
  - Added `stripe_cli` service for local webhook testing
  - Updated database port configuration
  - Added `.env.cli` and `.env.cli-test` for local development

## Usage Examples

Payment test data generation:
~~~bash
yarn backend-cli payment create-test-data --count 5 --method s_card --amount 2000 --period monthly --payFee --prorate
~~~

List API keys:
~~~bash
yarn backend-cli api-key list
~~~

Configure system:
~~~bash
yarn backend-cli configure --emailDomain example.com --paymentMethods stripe,gocardless
~~~

Sync segments with dry run:
~~~bash
yarn backend-cli sync segments --dryRun
~~~

## Changes
- Added new CLI commands
  - Payment test data generation
  - API key listing and creation functionality
  - System configuration command
  - Sync commands for Mailchimp and segments
  - Gift processing command
- Enhanced environment handling
  - Added Docker environment detection
  - Added environment-specific configuration files
    - `.env.cli` for local development
    - `.env.cli-test` for local testing
  - Updated database port configuration in Docker setup
- Improved code structure
  - Uses existing services (PaymentService, PaymentFlowService, ContactsService)
  - Maintains consistent URL configuration
  - Better error handling in API key commands
  - Added dry-run support for sync operations
  - Detailed logging for sync operations

## Deprecated Components
- Deprecated tools in `apps/backend/src/tools`:
  - `configure.ts` (use `backend-cli configure` instead)
  - `mailchimp/sync.ts` (use `backend-cli sync mailchimp` instead)
  - `new-user` (use `backend-cli user create` instead)
  - `start-gifts` (use `backend-cli process gifts` instead)
  - Ported and removed `resync-stripe.ts`
  - Ported and removed `list-users.ts`

## Testing
- [x] Verified contact creation and reuse
- [x] Tested payment flow creation
- [x] Verified API key listing functionality
- [x] Tested API key creation
- [x] Tested system configuration updates
- [x] Tested Mailchimp sync with dry-run
- [x] Tested gift processing
- [x] Tested CLI execution in Docker environment
- [x] Tested CLI execution locally


packages/core/src/providers/payment-flow/StripeFlowProvider.ts
```diff
@@ -89,7 +89,8 @@ class StripeFlowProvider implements PaymentFlowProvider {
 
     return {
       joinForm: { ...joinFlow.joinForm, paymentMethod },
-      customerId: "", // Not needed
+      customerId:
```
wpf500, line 92: I know casting isn't always the best option, but in this case I think it's better as it's how the Stripe docs recommend to do it. `setupIntent.customer` will always be a string here as we don't request for it to be expanded it in the Stripe API call

```ts
customerId: setupIntent.customer as string
```

packages/core/src/services/PaymentFlowService.ts
```diff
@@ -93,7 +93,7 @@ class PaymentFlowService implements PaymentFlowProvider {
     urls: CompleteUrls,
     completeUrl: string,
     user: { email: string; firstname?: string; lastname?: string }
-  ): Promise<PaymentFlowParams> {
+  ): Promise<PaymentFlow> {
```
wpf500, line 96: This change of signature needs to be reflected in the controllers. It's bad that this doesn't cause a type error, it really should. The relevant code is here

https://github.com/beabee-communityrm/monorepo/blob/163690db649eb276b03481c833382c141fe67658/apps/backend/src/api/controllers/SignupController.ts#L44-L55

and

https://github.com/beabee-communityrm/monorepo/blob/163690db649eb276b03481c833382c141fe67658/apps/backend/src/api/controllers/ContactController.ts#L387-L404

It seems `plainToInstance` isn't type safe, it takes a random object and tries to convert it to the given class. Long term we should review where we use this method and if there's a way to add more type safety. For now I would just fix the problem, e.g.

```ts
return plainToInstance(GetPaymentFlowDto, flow.params);
```

wpf500 (changes requested):
At least the payment flow problem needs to be fixed, otherwise it breaks the sign up flow

wpf500:
Really great to see the start of a decent CLI and much cleaner place to put our tools :tada:

packages/core/src/providers/payment-flow/StripeFlowProvider.ts
JumpLink: Is `setupIntent.customer?.toString() || ""` also okay for you to be type safe?

JumpLink: or that we throw an error if it's not a string? Personally, I prefer it when the code does something unnecessary than to work around the types.

packages/core/src/services/PaymentFlowService.ts
JumpLink: Oh okay, I thought the types were correct, but ‘GetPaymentFlowDto’ actually has the signature of ‘PaymentFlowParams’, the missing ‘Param’ confused me here.

packages/core/src/providers/payment-flow/StripeFlowProvider.ts
wpf500: I have changed it to at least check for the `null` condition, as I agree using `!` is always terrible. But I think casting to `string | null` is fine as you specifically have to ask Stripe to expand the object for it to be something else

wpf500 (approved):

PR #138, opened 2025-02-03
feat: add email for failed initial contributions

wpf500:
This PR adds an email for when a user's first payment fails on their contribution.

Waiting for email copy.

PR #139, opened 2025-02-03
Remove unnecessary add tag, now handled inside NewsletterService

wpf500:
Removes a now unnecessary active member tag check. This was made redundant by #124, as the `NewsletterService` and providers now always handles this internally.


JumpLink (approved):

wpf500:
I've rebased this so it doesn't depend on #138

PR #147, opened 2025-02-17
feat: add NL translations

wpf500:
Add Dutch translations and move all the translations to their own package (`@beabee/locale`) so they can be shared by the frontend and backend.

---
This is **part 1 of 2 in a stack**:
- <kbd>&nbsp;2&nbsp;</kbd> #148 
- <kbd>&nbsp;1&nbsp;</kbd> #147 👈


wpf500:
@JumpLink I tried to do a good thing and clean up some tech debt while I added the Dutch translations, but I got lost in what I suspect is a CommonJS/ESM hole :sob:.

Currently it's building but doesn't actually work. I really don't understand what the problem is but maybe you can see it right away!

JumpLink (dismissed):

wpf500:
@JumpLink not sure why the tests are failing here, it says the API_KEY is missing again

JumpLink:
@wpf500 To see why the test failed, I noticed a few things in the E2E test that still needed to be standardised in order to standardise the env variable parsing in the tests. I did this with #150 in a separate PR. This was possibly part of the problem why the test failed, but the main problem was still a faulty configuration of the building process of the locale package, which should have been fixed with my last commit. If the test continues to fail, I recommend first merging #150 into main and then merging main into this branch; this should then finally fix the problem.

JumpLink (approved):

PR #149, opened 2025-02-24
feat: add query parameter for map center and zoom level

wpf500:
Add query parameters to track the zoom and center level. This is inline with what most map tools do (Google Maps, Bing, OpenStreetMap, etc.), it enables users to share particular positions on a map, and for our use case in particular enables admins to share links with the map positioned in a different place on page load.


JumpLink (approved):
LGTM

PR #152, opened 2025-02-26
refactor(locale): Use unquoted keys if possible + branch name

JumpLink:
Since the translations are now generated in `.js` format instead of `.json` format, the keys no longer need to contain quotation marks. This reduces the size of the translation files and makes them more readable.

In addition, the script is now passed the current branch name again.


wpf500:
I think it would be better to export the locales as JSON again as this will be required if we want to use a localisation platform like Weblate. So I'm not sure if it's worth merging this just to undo it again later? What do you think?

JumpLink:
> I think it would be better to export the locales as JSON again as this will be required if we want to use a localisation platform like Weblate. So I'm not sure if it's worth merging this just to undo it again later? What do you think?

You mean for the import in Weblate? A simple script would suffice:

```js
import {de} from 'de.js';
console.log(JSON.stringify(de));
```

I think the effort would be negligible, but if you prefer, I can undo that part in this PR so that the keys remain in quotation marks.

wpf500:
Weblate works by having direct access to the JSON files that it edits and commits changes to. It supports [lots of formats](https://docs.weblate.org/en/latest/formats.html) of course, but it understandably can't read a TypeScript file.

So to support Weblate we would need to put the translations back into a JSON file, then maybe use a simple TS file to expose them in the package, e.g.
```ts
import en from './en.json';
export { en }
```

I think it would make most sense to just keep the branch part of this PR, and drop the file formatting

JumpLink:
okay, then maybe I can do it the way you suggested instead, so that they are json files again.

wpf500:
If you are happy to do this then feel free :slightly_smiling_face:, otherwise I was thinking I might have a go at some point in this sprint. I'm going to look at the locale stuff now so will separate out the i18n branch support from this branch

JumpLink:
@wpf500 thank you

PR #154, opened 2025-02-27
chore: generate locale from branch name

wpf500:
Generate locales for a branch, ported from #152

### DIFF
--- a/packages/locale/generate.ts
+++ b/packages/locale/generate.ts
@@ -1,7 +1,8 @@
 /**
  * Script to load locale data from a Google sheet
  *
- * Run: ./i18n.js [sheet name]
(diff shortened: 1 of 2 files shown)


JumpLink (approved):
Maybe you can also update the script comment for this:

```js
/**
 * Script to load locale data from a Google sheet
 *
 * Run: `yarn generate` to generate a locale for the current branch.
 * Run: `yarn generate:branch [sheet name]` to generate a locale for a branch of your choice.
 *
 * Column headers should be locale codes, codes that start with an exclamation mark (!) are ignored.
 *
 * The script can optional load a second sheet to overwrite the main sheet, we add a new sheet for a
 * branch so changes for different features are kept separate.
 */
 ```
 
 But since the script will probably be removed soon anyway, it's not that important.

PR #155, opened 2025-02-27
fix: language switching

wpf500:
The recent changes to introduce `@packages/locale` broke the internationalisations. It turns out `vue-i18n` only supports message compilation from JSON files, so this PR puts the locales in JSON files again. This will also be helpful when migrating to Weblate


packages/locale/esbuild.js
```diff
@@ -29,6 +29,19 @@ await build({
   format: "cjs",
 });
 
+// The JSON locale files need to be available to both module formats, but also
+// in the types directory so they can be used to generate static types
+for (const outdir of [OUTDIR_ESM, OUTDIR_CJS, './dist/types']) {
+  await build({
+    entryPoints: ["./src/**/*.json"],
+    outdir,
+    bundle: false,
```
JumpLink, line 38: Maybe you can minify it here?

JumpLink (approved):
LGTM, Just a small optimisation suggestion

packages/locale/esbuild.js
wpf500: I had a go but it looks like the `minify` option is bypassed with the `copy` loader, unless there is another way to do it? In any case these files are compiled by `vue-i18n` in the frontend so minification happens there.

JumpLink: but they are loaded from the server via an async import, right? Then they wouldn't be minified yet, would they?

Instead of using esbuild, you could just do something like this:

```ts
import fs from 'fs'
import path from 'path'
import glob from 'glob'

// Find all JSON files
const jsonFiles = glob.sync('src/**/*.json')

// Process each file
jsonFiles.forEach(file => {
  const json = fs.readFileSync(file, 'utf8')
  const minified = JSON.stringify(JSON.parse(json))
  
  // Create output path
  const outputPath = path.join('dist', file.replace('src/', ''))
  fs.mkdirSync(path.dirname(outputPath), { recursive: true })
  
  // Write minified JSON
  fs.writeFileSync(outputPath, minified)
})
```

But we don't need to put any effort into it now

wpf500: I'm not sure how it works to be honest but fairly sure `vue-i18n` does some Vite magic some how :shrug: . I've merged this for now but will have a look next week.

wpf500: I don't really get how it's figuring it out... but here is an example of a minified file on production :)

https://demo.beabee.io/assets/nl-Df2M2LlJ.js

PR #159, opened 2025-03-26
feat: create contact from guest submission

wpf500:


apps/backend/src/api/dto/CalloutResponseDto.ts
```diff
@@ -109,18 +109,26 @@ export class GetCalloutResponseDto {
...
+export class CreateCalloutResponseGuestDto {
+  @IsString()
+  firstname!: string;
+
+  @IsString()
+  lastname!: string;
+
+  @IsEmail()
+  email!: string;
+}
```
JumpLink, line 121: Is it possible to check here that the strings are not empty?

apps/backend/src/api/controllers/CalloutController.ts
```diff
@@ -175,7 +175,7 @@ export class CalloutController {
       throw new NotFoundError();
     }
 
-    if (caller && (data.guestEmail || data.guestName)) {
+    if (caller && data.guest) {
```
JumpLink, line 178: Only works this way if we can be sure that `data.guest` is `undefined` or `null` if no guest is set, otherwise we should also check the defined properties here to be not an empty string like `if (caller && data.guest && data.guest.email)`

apps/frontend/src/components/pages/callouts/CalloutForm.vue
```diff
@@ -96,8 +97,11 @@ const props = defineProps<{
   onSubmit?(answers: CalloutResponseAnswersSlide): void;
 }>();
 
-const guestName = ref('');
-const guestEmail = ref('');
+const guestData = reactive({
+  firstname: '',
+  lastname: '',
+  email: '',
+});
```
JumpLink, line 104: empty strings are probably fine here, but I'm commenting on it anyway because it's relevant to my other comments

apps/frontend/src/components/pages/callouts/CalloutForm.vue
```diff
@@ -173,12 +177,9 @@ async function handleSubmit() {
...
       props.callout.slug,
       {
-        ...(!currentUser.value &&
-          props.callout?.access === 'guest' && {
-            guestName: guestName.value,
-            guestEmail: guestEmail.value,
-          }),
         answers: validAnswers,
+        ...(!currentUser.value &&
+          props.callout?.access === 'guest' && { guest: guestData }),
```
JumpLink, line 182: Maybe we should also check for empty strings here?

packages/client/src/api/callout.client.ts
```diff
@@ -177,19 +177,12 @@ export class CalloutClient extends BaseClient {
...
     captchaToken?: string
   ): Promise<GetCalloutResponseData> {
     const { data } = await this.fetch.post<Serial<GetCalloutResponseData>>(
       `/${slug}/responses`,
-      {
-        answers: newData.answers,
-        guestName: newData.guestName,
-        guestEmail: newData.guestEmail
-      },
+      { answers: newData.answers, guest: newData.guest },
```
JumpLink, line 185: Should the backend throw en error for empty strings here or should we also catch this in the client directly?

packages/core/src/services/CalloutsService.ts
```diff
@@ -266,29 +268,29 @@ class CalloutsService {
    */
   async setGuestResponse(
     callout: Callout,
-    guestName: string | undefined,
-    guestEmail: string | undefined,
+    guest: CreateCalloutResponseGuestData | undefined,
     answers: CalloutResponseAnswersSlide
   ): Promise<CalloutResponse> {
-    if (callout.access === CalloutAccess.Guest && !(guestName && guestEmail)) {
+    if (callout.access === CalloutAccess.Guest && !guest) {
```
JumpLink, line 274: Also empty string relevant here?

JumpLink (commented):
It mostly looks good, I'm just wondering if it could be a problem to only check the guest object but not to check whether the properties of it are also set to ensure that they are not empty strings (which are the default values in the frontend).

apps/backend/src/api/dto/CalloutResponseDto.ts
wpf500: Hmm good point! `IsEmail` covers this because an empty string isn't a valid email address, but `IsString` would need checking further. I guess this is probably a common problem across a lot of our validation

apps/backend/src/api/controllers/CalloutController.ts
wpf500: I will resolve the rest of these comments as I think the most logical place to check this is in the DTO validation

wpf500:
@JumpLink I've added a check for non-empty strings on the first and last name. This validator should probably be used in lots of places where we currently use `IsString`, but this should be addressed in a different PR

PR #160, opened 2025-03-26
feat: show guest NL opt in

wpf500:
This PR adds a newsletter opt in flow when users are responding to callouts

### To do
- [x] Fix reactivity bug with newsletter groups (`AppRepeatable`)
- [x] Add newsletter signup warning
- [x] Decide on the copy
- [x] Internationalise strings

### Assumptions
- Selected newsletter groups are merged with the contact's existing newsletter groups, not overwritten. So if a contact was already subscribed to a group but didn't tick it in the callout, they won't be removed from that group.
- Contacts are created for all submissions with guest information (not just when it's required)

### Screenshots

Admin settings
grafik

User view
grafik

### Tech debt

This PR encounters tech debt caused by how newsletter status/groups are currently handled. As a result, this PR introduces new complexity to `ContactsService.updateContactProfile` to support merging newsletter groups.

The `NewsletterService` currently has a dependency on the `ContactProfile` model which creates poor separation between the `ContactsService` and `NewsletterService`.  To fix this:

* Newsletter fields (`newsletterStatus` and `newsletterGroups`) should be put in a new model (e.g. `ContactNewsletter`)
* `ContactsService.updateContactProfile` should no longer handle any newsletter related fields
* A new method (e.g. `ContactsService.updateContactNewsletter`) should handle then instead.
* `NewsletterService` should remove all references or dependencies on `ContactProfile` or `contact.profile`
* Add a new route (`/contact/:id/newsletter`) to handle newsletter updates
  * Consider handling `groups` array updates in the same way as `tags` array updates are handled for consistency across the API, i.e. using the `+`/`-` syntax
    ```
    PATCH /contact/:id/newsletter
    {
      "status": "subscribed",
      "groups": ["+group1", "-group2"]
    }
    ```
* The route `/contact/:id/profile` should just handle profile updates


JumpLink (commented):
It looks good to me so far ⭐

Regarding the technical debt mentioned around newsletter handling and the separation between `ContactsService` and `NewsletterService`, how do you plan to address this? 
Would you prefer to tackle these improvements in a separate PR? If so, I'd recommend creating a GitHub issue that documents these problems and the proposed solutions (as outlined in your PR description) to ensure they don't get lost in the backlog. 

And what about the other two missing TODO's?

- [ ] Decide on the copy
- [ ] Internationalise strings

wpf500:
I would address the tech debt separately, I'm just trying to get a bit into the process of documenting it somewhere. I will create a GitHub issue as you suggest :+1:

I've just made up some copy and didn't put much thought into it, the todos are just to ensure it gets thought about before we merge. I have talked to Sara about reviewing it when she takes a look (the feature is already on dev), then once we've agreed the final wording we just need to internationalise as usual.

wpf500:
@JumpLink ready for final review I think. I've fixed a bug with newsletter status not syncing and added German translations.

wpf500:
Merging for pre-release, still needs final review

PR #146, opened 2025-02-13
Feat: Callout admin UX improvements

JumpLink:
The goal of this pull request is to make the callout admin UX more modern and intuitive based on the new designs:

* [Setting options and defaults](https://github.com/user-attachments/files/18779902/Setting.options.and.defaults.pdf)
* [Error notices](https://github.com/user-attachments/files/18779901/Error.notices.pdf)
* [Edit UI](https://github.com/user-attachments/files/18779899/Edit.UI.pdf)
* [Drafts](https://github.com/user-attachments/files/18779897/Drafts.pdf)

PR #162, opened 2025-04-01
refactor: new components in @beabee/vue

wpf500:
Move lots of components to `@beabee/vue`. This PR extracts the generic changes out of #146


apps/frontend/src/components/forms/AppInput.vue
```diff
@@ -260,4 +299,8 @@ function isPassword(value: string) {
 watch(validation, (newState) => {
   emit('update:validation', !newState.$invalid);
 });
+
+const handleCopy = () => {
```
wpf500, line 303: I think we should remove stub methods like this

packages/vue/src/components/scroll/AppScrollNavigation.vue
```diff
@@ -0,0 +1,144 @@
...
+  if (sectionsWithElements.length === 0) return;
+
+  // Find the section that is currently in view
+  // We use a small offset to improve the detection
+  const offset = 100;
+
+  for (const section of sectionsWithElements) {
+    if (!section.element) continue;
+
+    const rect = section.element.getBoundingClientRect();
```
wpf500, line 97: I think this is a really expensive operation to put in an event listener, oder? I'd like to take another look at the scroll components, but I will do it in a separate PR

packages/vue/src/components/scroll/AppScrollNavigation.vue
```diff
@@ -0,0 +1,144 @@
...
+    if (rect.top <= offset && rect.bottom > offset) {
+      if (activeSection.value !== section.id) {
+        activeSection.value = section.id;
+        emit('update:activeSection', section.id);
+        emit('section-change', section.id);
```
wpf500, line 102: Why are there two events for the same thing?

apps/frontend/src/components/forms/AppInput.vue
JumpLink: yes

packages/vue/src/lib/theme.ts
```diff
@@ -136,6 +136,46 @@ export function getFullTheme(theme: PartialTheme): Theme {
...
 
+/**
+ * Dynamically loads font CSS files as separate chunks.
+ * By using static import paths, Vite can properly analyze and create separate chunks
+ * for each font style during build time, enabling efficient code-splitting and lazy loading.
+ *
+ * @param font - The font name to load
+ * @returns A promise that resolves when the font CSS is loaded
+ */
+function loadFont(font: string) {
```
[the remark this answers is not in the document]
JumpLink: I had a warning about this in vite and therefore changed it

packages/vue/src/types/form.interface.ts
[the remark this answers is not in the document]
JumpLink: Maybe yes, so far we had the rule to put types and interfaces in the `types` folder, so I did the same here. I found it difficult to find a good structure for the components. I created a props interface for each component and left it in the component because it is an interface that is closely related to the component.

We should also find a rule on how we want to do it here. Within the frontend, it would be consistent to do it as you suggested, but on the other hand, we do it in the other packages as described. A `types` folder also has the advantage that we can immediately see that we only need to import the type and that we always know where to look for types or where to put them. If we don't do it this way, we may have to decide each time what to call it and where to place it.

In the end, I don't care how we do it, I just think it's important to have an easy-to-follow rule and then to be consistent

docker-compose.yml
```diff
@@ -60,6 +60,7 @@ services:
       - ./packages:/opt/packages
       - ./apps/backend/built:/opt/apps/backend/built
       - ./apps/backend-cli/dist:/opt/apps/backend-cli/dist
+      - ./packages/locale/dist:/opt/packages/locale/dist
```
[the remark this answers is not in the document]
JumpLink: Yes, that's right, it was probably an accident.

packages/vue/package.json
```diff
@@ -64,14 +68,17 @@
     "./components/button/*": "./src/components/button/*.vue",
```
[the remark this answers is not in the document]
JumpLink: Yes, that's how I've usually imported them so far and it hasn't caused any problems

JumpLink:
Do you want to apply the changes or should I do it?

packages/vue/src/components/scroll/AppScrollNavigation.vue
JumpLink: Yes, you're right, `getBoundingClientRect` is quite computationally intensive for the browser and it takes a lot of effort to execute it every time you scroll, so I agree that we should optimise it

packages/vue/src/components/scroll/AppScrollNavigation.vue
JumpLink: I changed the code several times over time and this is probably a leftover from that. Originally, I planned to do a complete cleanup of such things at the end.

packages/vue/src/types/form.interface.ts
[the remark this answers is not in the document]
wpf500: I've ended going the other way and moving `tabs.interface.ts` to the types folder

wpf500:
@JumpLink I'm merging this without addressing the `AppScroll*` component problems as this can be resolved in #164

PR #163, opened 2025-04-01
feat(s3): Add MinIO service and configuration for file storage

JumpLink:
- Introduced MinIO service in docker-compose.yml for object storage.
- Added MinIO client to initialize buckets and set policies.
- Updated .env.example with MinIO configuration variables.
- Modified docker-compose.test.yml to include MinIO service dependencies and environment settings for testing.
- Enhanced image upload app configuration to utilize MinIO for S3 storage.

- [ ] Done a self-review of my code

## Checklist before merging


JumpLink:
@wpf500 Uploading to the s3 storage is already working, so the biggest problem now is likely to be the migration of the old images.

JumpLink:
Ready for review, as this is the base PR for #175, it does not need to be tested further and should instead be tested together with #175.

wpf500:
@JumpLink I've just reviewed `feat/image-service-2`. Can this branch just be merged into that one? These changes look like they are then being changed again in that branch, so I'm not sure if it makes sense for me to review this separately?

wpf500:
Note as this makes changes to the Docker Compose stack we will also need equivalent `docker-compose.yml` file updates in:

- https://github.com/beabee-communityrm/hive-deploy-stack
- https://github.com/beabee-communityrm/beabee-deploy

JumpLink:
> @JumpLink I've just reviewed #175. Can this branch just be merged into that one? These changes look like they are then being changed again in that branch, so I'm not sure if it makes sense for me to review this separately?

Yes, the other branch replaces some stuff from here, so de can merge it to make it easier to review

docker-compose.yml
```diff
@@ -88,13 +88,29 @@ services:
...
+      context: .
+      dockerfile: ./apps/minio/Dockerfile
+    ports:
+      - "${BEABEE_MINIO_PORT_ADMIN:-9000}:9000"  # Admin port for debugging, not for production
+      - "${BEABEE_MINIO_PORT_CONSOLE:-9001}:9001"  # Console port (Admin UI), not for production
     volumes:
-      - upload_data:/var/www/data
+      - minio_data:/data
+    healthcheck:
+      test: ["CMD", "curl", "-f", "http://localhost:9000/minio/health/live"]
```
wpf500, line 102: I think it might also be possible to move the healthcheck config into the Dockerfile

docker-compose.test.yml
```diff
@@ -45,12 +45,19 @@ services:
...
     extends:
       file: docker-compose.yml
-      service: img_upload_app
+      service: minio
+    ports:
+      - "${BEABEE_MINIO_PORT_ADMIN:-9002}:9000" # Admin port for debugging, not for production
+      - "${BEABEE_MINIO_PORT_CONSOLE:-9003}:9001" # Console port (Admin UI), not for production
     environment:
-      URL: ${BEABEE_AUDIENCE}/uploads/
+      MINIO_ROOT_USER: ${BEABEE_MINIO_ROOT_USER:-minioadmin}
```
wpf500, line 56: These env vars are duplicated from `docker-compose.yml`, are they needed?

packages/core/src/models/UploadFlow.ts
```diff
@@ -1,28 +0,0 @@
-import {
```
wpf500, line 1: Needs a migration to delete the table

wpf500 (changes requested):
I'm interested in what your thoughts are about the try/catch block stuff!

Otherwise the only required change is to add a migration for the `UploadFlow`, otherwise this could leave dangling foreign keys to the contacts table (would cause errors when trying to delete those contacts)

packages/core/src/services/DocumentService.ts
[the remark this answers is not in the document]
JumpLink: You're right, we really only need to catch the multer errors; everything from our own services should already return the correct error types (based on HttpError). I've simplified it considerably now.

apps/frontend/src/utils/url.ts
```diff
@@ -0,0 +1,26 @@
...
+/**
+ * Resolves an image URL to ensure it's absolute.
+ * If the provided URL is relative (and not a blob URL), it prepends the API base URL.
+ * Absolute URLs, blob URLs, or URLs already pointing to the API are returned as is.
+ * Returns an empty string if the input is not a valid string.
+ *
+ * @param url - The image URL to process (can be relative or absolute).
+ * @param width - The width of the image to be resolved.
+ * @returns The resolved, absolute image URL or an empty string.
+ */
```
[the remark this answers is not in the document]
JumpLink: Is string or URL now

docker-compose.test.yml
JumpLink: If they are not needed, it means that other environment variables in this file can also be removed 🤔

docker-compose.yml
JumpLink: Thank you, works

docker-compose.test.yml
JumpLink: Thank you, works

JumpLink:
@wpf500 Everything implemented and [deployed on dev](https://dev.beabee.io), including minio health check, which is now part of the docker image.

wpf500 (dismissed):
Great work :tada:

wpf500:
Ah I just noticed that the CNR references hadn't been fixed, I've done that in [this commit](https://github.com/beabee-communityrm/monorepo/commit/7fd207e1e6ee05ac7c4d7d4405a787e926a3881d)

wpf500 (approved):

PR #164, opened 2025-04-02
feat: callout admin UX changes (rebased)

wpf500:
This PR supersedes #146 as the generic changes have been separated out into #162 

The goal of this pull request is to make the callout admin UX more modern and intuitive based on the new designs created in [Penpot](https://design.penpot.app/#/workspace/887044a0-9c39-11eb-b4cd-81aaa3a367f9/ee63301e-1fa7-81b1-8004-64c00becb673?page-id=ee63301e-1fa7-81b1-8004-64c00becb674) :

* [Setting options and defaults](https://github.com/user-attachments/files/18779902/Setting.options.and.defaults.pdf)
* [Error notices](https://github.com/user-attachments/files/18779901/Error.notices.pdf)
* [Edit UI](https://github.com/user-attachments/files/18779899/Edit.UI.pdf)
* [Drafts](https://github.com/user-attachments/files/18779897/Drafts.pdf)
* [Structure](https://github.com/user-attachments/files/18843803/Callout.admin.UX.improvements_2025-02-18_11-14-06.pdf) ([Mural](https://app.mural.co/t/beabee6721/m/beabee6721/1739872401901/e7e5a4995244a0758be90d42ef5dabdb26550a30))

### Todo
- [ ] Test embeds
- [x] Fix autoSlug
- [x] Remove active scroll section highlight
- [x] Remove advanced options toggle

### Todo from Marc from [Trello](https://trello.com/c/238eTFTu)
- [x] **Share settings reversed logic**: Under "Title and Image" tab, users must deactivate "Use callout title and description for sharing?" to use the text from above - the logic should be reversed
- [x] **Display answers persistence bug**: When "Display answers publicly" is enabled and then later disabled, the card settings remain visible instead of disappearing
- [ ] **Publication blocked**: Unable to publish the callout despite all tabs showing checkmarks (may be related to the above issues)

### Out of scope to do

All existing problems related to FormIO that shouldn't block this feature

- [ ] **Inconsistent language**: Elements labeled "Basic/Advanced" remain in English while the rest of the UI is in German
- [ ] **Address field issues**: Address field requires a provider to be entered without clear instructions, making it difficult for users to complete without assistance
- [ ] **Phone number format limitation**: Phone number validation appears to only accept British format, limiting international usability

Can release with current implementation
- [ ] Rework scrollable section components


apps/frontend/src/components/rte/RichTextEditor.vue
```diff
@@ -165,6 +195,10 @@ function setLink() {
       .run();
   }
 }
+
+function handleCopy() {
```
wpf500, line 199: Remove stub method

apps/frontend/src/layouts/menu/TheMenuList.vue
```diff
@@ -1,5 +1,7 @@
 <template>
-  <div class="flex-auto">
+  <div
+    class="mb-[53px] flex-auto overflow-y-auto overflow-x-visible lg:mb-[61px]"
```
wpf500, line 3: Where do `53px` and `61px` come from? Could this be made explicit somehow?

apps/frontend/src/utils/callouts.ts
```diff
@@ -262,70 +355,87 @@ function convertVariantsForCallout(
...
-          imageProp: steps.settings.responseImageProp,
-          imageFilter: steps.settings.responseImageFilter,
-          gallery: steps.settings.responseViews.includes('gallery'),
-          links: steps.settings.responseLinks,
-          map: steps.settings.responseViews.includes('map')
+          buckets: tabs.responseDisplay.responseBuckets,
+          titleProp: tabs.responseDisplay.responseTitleProp,
+          imageProp: tabs.responseDisplay.responseImageProp,
+          // TODO: Remove this once the image filter is removed from the API
+          imageFilter: '',
```
wpf500, line 404: This needs the preserve the current callout `imageFilter` value somehow to not overwrite it if it already exists

apps/frontend/tsconfig.json
```diff
@@ -25,7 +25,14 @@
       "@assets/*": ["./src/assets/*"]
     }
   },
-  "include": ["src/**/*.ts", "src/**/*.d.ts", "src/**/*.tsx", "src/**/*.vue"],
+  "include": [
+    "src/**/*.ts",
+    "src/**/*.d.ts",
+    "src/**/*.tsx",
+    "src/**/*.vue",
+    "../../packages/vue/src/plugins/icons.ts",
```
wpf500, line 33: Can you explain why this is necessary?

packages/vue/src/components/tabs/AppTabCard.story.vue
```diff
@@ -90,8 +90,8 @@ const longContent = Array(20)
             variant="boxed"
           >
             <template #default="{ selected }">
-              <div class="space-y-4">
-                <AppFormBox title="Section 1">
+              <div class="py-4">
+                <AppFormBox class="!px-0" title="Section 1">
```
wpf500, line 94: Do we really need an `!important` class here?

packages/vue/src/components/tabs/AppTabCard.vue
```diff
@@ -2,26 +2,23 @@
...
-      'bg-white px-1 pt-1': variant === 'boxed',
-      'bg-primary-5': variant === 'transparent',
-      'rounded-t-lg border border-white': variant === 'transparent',
     }"
   >
     <AppTabs
       :items="items"
       :selected="modelValue"
-      class="!mb-0 !rounded-none !border-0 !p-0"
+      class="!mb-0 !rounded-t-lg !p-0"
```
wpf500, line 13: Do we really need `!important` classes here?

apps/frontend/src/layouts/layout-Dashboard.vue
```diff
@@ -1,22 +1,35 @@
 <template>
-  <div v-if="isEmbed" class="flex justify-center">
+  <div v-if="isEmbed" class="flex h-screen w-full flex-col md:flex-row">
```
wpf500, line 2: Need to test some embeds here I think, not sure I understand why these classes are being applied

apps/frontend/src/pages/admin/settings/theme.vue
```diff
@@ -45,40 +45,43 @@ meta:
...
-          />
-          <span class="ml-2 flex-1">
-            {{ t('adminSettings.theme.presetColors.' + preset.name) }}
-          </span>
-          <span
-            v-for="(color, name) in preset.colors"
-            :key="name"
-            class="ml-2 h-8 w-8 rounded-full"
-            :style="{ backgroundColor: color }"
+            :wrapper-class="'mb-2 flex items-center'"
```
wpf500, line 57: Simpler without interpolation

```
wrapper-class="mb-2 flex items-center"
label-class="w-full"
```

apps/frontend/src/pages/admin/settings/theme.vue
```diff
@@ -45,40 +45,43 @@ meta:
...
+          <AppRadioInput
             v-model="activePreset"
-            type="radio"
             name="activePreset"
             value="custom"
-            class="mr-2"
-          />
-          {{ t('adminSettings.theme.customColors') }}
-        </label>
+            :wrapper-class="'mb-4 flex h-8 items-center'"
```
wpf500, line 80: Simpler without interpolation

packages/vue/src/components/scroll/AppScrollNavigation.story.vue
```diff
@@ -14,11 +14,6 @@ const sections: ScrollSection[] = [
     id: 'section4',
     label: 'Section 4 with a very long name that should be truncated',
   },
-  {
```
wpf500, line 17: Not sure why this was deleted, maybe a merge problem?

apps/frontend/src/components/pages/admin/callouts/tabs/SettingsTab.vue
```diff
@@ -0,0 +1,443 @@
...
+
+const canStartNow = computed(
+  () =>
+    !props.status ||
+    props.status === ItemStatus.Draft ||
+    props.status === ItemStatus.Scheduled
+);
+
+// Computed property to control start date toggle
+const startDateEnabled = computed({
```
wpf500, line 351: Do we need this proxy property? Couldn't the `v-model` just use `localData.startNow` directly?

apps/frontend/src/components/pages/admin/callouts/tabs/TranslationsTab.vue
```diff
@@ -0,0 +1,995 @@
...
+                </AppFormBox>
+              </AppScrollSection>
+            </template>
+          </AppTabCard>
+        </div>
+      </div>
+
+      <!-- Right Sidebar -->
+      <div class="flex-0 basis-[15rem] overflow-y-auto">
+        <!-- TODO: Add right sidebar content here -->
```
wpf500, line 553: Remove TODO?

wpf500 (commented):
I have a few comments here and there, but in general this is really impressive implementation :tada:.

I'd like to address a few things, and am mostly leaving comments to remind myself but would appreciate it if you could look at some of the questions.

One thing in particular I'm not sure about is the implementation of the scrollable sections. It has a lot of boilerplate (e.g. storing element refs) and using `getBoundingClientRect` within a scroll handler is generally bad for performance (even on my laptop is causing problems). I will have a look at this

apps/frontend/src/components/pages/admin/callouts/tabs/TranslationsTab.vue
JumpLink: Yes, TODO can be removed, but I would leave the comment to make it clear that this is the space for the content of an optional right sidebar that is uniform with the right sidebar of the content editor.

apps/frontend/src/layouts/menu/TheMenuList.vue
JumpLink: By manually fine-tuning, but I'm happy to check again whether one of the standard values can be used here too.

JumpLink: I remember manually adjusting the spacing in the menu to get everything aligned properly, but you're welcome to double-check if it's possible with the default values.

apps/frontend/src/components/pages/admin/callouts/tabs/SettingsTab.vue
JumpLink: Yes, it looks like it

apps/frontend/src/components/forms/AppTextArea.vue
```diff
@@ -1,27 +1,70 @@
...
+        :required="required"
+        :disabled="disabled"
+        :aria-invalid="hasError"
+        :aria-describedby="getAriaDescribedBy"
+        v-bind="$attrs"
+        @blur="validation.$touch"
+      />
+      <div v-if="copyable" class="absolute right-0 top-0">
+        <AppCopyButton
+          class="mr-1 mt-1 !h-9 !w-9 rounded bg-white/70 hover:bg-white"
```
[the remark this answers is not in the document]
JumpLink: Yes, I had already given the button a size and overwriting it is a bit ugly, I agree. Maybe just pass the desired size as a property and set it dynamically? The width and height should always be the same, so only one size property would be needed for that.

apps/frontend/tsconfig.json
JumpLink: That doesn't belong there and probably causes problems. My VSCode likes to enter that automatically when I move files, so I should switch it off

packages/vue/src/components/tabs/AppTabCard.story.vue
JumpLink: At least it looks like the div wrapper could simply be removed and then `!px-0` could be omitted.

JumpLink:
grafik

Unfortunately, the automatic generation of the autoSlug does not work, it always remains empty and I have not been able to fix it. Actually, this has worked before, originally I had solved it like this:
* If autoGenerateSlug active, set the autoSlug based on the title
* If autoGenerateSlug inactive, then set customSlug only if it is still empty
* In both cases, only change the slug if `canEditSlug` is `true

I also tried the following, but it didn't help either:
* When `autoGenerateSlug` is changed, update the corresponding slug according to the rules above.

But for some reason this no longer works. I tried to rewrite it several times, but none of it helped, which is why I didn't push it in the end

wpf500:
@JumpLink I've made quite a few changes!

* Fixed the slug generation problem
* Fixed map settings not hiding properly
* Split `TranslationTab.vue` into multiple components to try to understand what was going on and improve readability
* Moved all the computed logic in `SettingsTab.vue` into `convertCalloutToTabs` and `convertTabsToCallout`. In general the idea of these methods was to translate between the API schema and the frontend schema, so the frontend schema can directly reflect what is in the tabs without an additional layer of translation within the tabs.
* Removed `localData`. We were using a combination of manipulating props directly and using a `localData` proxy, I've switched to the former for consistency. The tab components are tightly connected to their parent, so this is a case where I think manipulating props is okay and more readable.
  * I've left `localData` in `ResponseDisplayTab.vue` because that whole file will be changed when we implement a cleaner UI for map settings
* Removed the important styles wherever I could find them. I'd like to add an ESLint rule to guard against important styles, we should never really use them!
* Moved some tab properties to the more relevant tab (e.g. `settings.locales` should now be in `translations.locales`)

wpf500:
I suggest that we don't rework the scroll section code in this PR so as to not block the release, but put something in the backlog to clean it up

JumpLink (approved):
Originally, my idea of the `AppTabCard` was that it should look like a component that belongs together, so that it is clear what belongs to the tabs and what does not. Since there is now a transparent background and the navigation at the bottom is rounded again, this is no longer the case.

grafik
grafik

This is a visual and not a technical design decision and should be clarified again if necessary.

Otherwise, I have added missing German translations and corrected an incorrect German one.

wpf500:
I mostly removed the non-rounded borders (and a few other tweaks) to remove the `!important` styles as it felt more important to not have those and quicker than refactoring, but I agree we should revisit  these components :slightly_smiling_face:

PR #170, opened 2025-04-14
feat: add active user filter

wpf500:
Adds a filter which checks if the user has set a password on their account, which is how we define a user as being active.

Also fixes a bug with `AppRadioInput` where it wouldn't allow `false` as a valid value (via `value || null`). Using `defineModel` is clearer in any case and removes this problem.

```vue
<template>
  <AppRadioInput v-model="myVal" :value="false" /> <!-- should be ticked -->
  <AppRadioGroup
    v-model="myVal"
    :options="[
      [false, 'Selected no'], 
      [true, 'Selected yes']
    ]"
  /> <!-- should select "Selected no" -->
</template>

<script setup>
const myVal = ref(false)
</script>
```


JumpLink (dismissed):
LGTM 👍

PR #175, opened 2025-04-16
Feat(image): Image upload controller + service

JumpLink:
This PR introduces a comprehensive image and document handling service using MinIO for storage while keeping all access controlled through our API.

* This PR replaces #165 and #166, and also makes #177 redundant, but can be merged if this PR cannot be merged yet.

## Solutions and advantages

* Images are not scaled during upload but when they are first accessed. This has a major advantage:
  * Permitted image sizes can be adjusted at any time
  * No migration scripts are required for new scaled images
  * Image sizes that are permitted but not used are not generated
* A separate Docker image service is no longer necessary, which reduces complexity as our own image service is much simpler to set up and can be adapted to our own needs.
* Our own service supports PDFs and also performs basic validation of the PDF.
* Our own service can be easily adapted for other data formats.
* Only images and documents that are actually used are migrated; files that are no longer used are discarded.
* Deleting data can also be easily implemented and should be implemented in a further PR.
* More supported image formats: jpeg, png, webp, gif, avif and svg
* Avif is now the standard format that delivers the best results in terms of image size and quality and is now supported by all big browsers: Chrome, Edge, Firefox, Opera and Safari. 
* SVG images are automatically optimised and the size scale is ignored 

## Changes

- Implemented a fully integrated image service using MinIO as the storage backend
- Created a dedicated document service for PDF uploads and retrieval
- Added a controller-based approach for direct image and document streaming instead of external URLs
- Added support for automatic image resizing with multiple dimensions
- Enhanced image format conversion with AVIF as the default format
- Developed proper error handling for image and document processing and delivery
- Modified interceptors to correctly handle binary data responses
- Extended migration script to migrate existing uploads to the new storage system
- I have now added a `path` attribute for files in addition to `url` and marked `url` as deprecated. This should make future changes to file URLs easier; for example, the host domain can now be changed without any problems. 

## TODO
- [x] Use relative URL's instead of absolute (This makes the next changes easier)

## Testing

- [x] Verified image upload functionality with various image types
- [x] Tested automatic image resizing with different width parameters
- [x] Confirmed correct content types are set for different image formats
- [x] Verified document upload and retrieval functionality for PDF files
- [x] Test Logo upload + response (`ContentGeneral`)
- [x] Test share image upload + response ('ContentShare')
- [x] Test membership / sign in background image upload + response
- [x] Test PDF upload + response
- [x] Test migration of existing uploads to new storage system
- [x] Test migration of callout title  images, callout response images, logo and share image
- [x] Test migration of callout response pdf uploads
- [x] Test error messages for wrong file sizes and wrong file formats
- [x] Validated error handling for missing images, documents, and invalid requests
- [ ] Test on dev instance (Depends on hive-deploy-stack [PR](https://github.com/beabee-communityrm/hive-deploy-stack/pull/9))
- [x] Rate limits
- [x] Different rate limits for users and guests
- [x] Add fallback for old shared images like the logo in emails, e.g. a NGINX redirect to the old file in the volume if possible
- [x] Test fallback described above
- [x] Adapt `docker-compose.yml` on [hive-deploy-stack](https://github.com/beabee-communityrm/hive-deploy-stack) ([PR ready for review](https://github.com/beabee-communityrm/hive-deploy-stack/pull/9))
- [x] Adapt `docker-compose.yml` on [beabee-deploy](https://github.com/beabee-communityrm/beabee-deploy) ([PR ready for review](https://github.com/beabee-communityrm/beabee-deploy/pull/20))

## Breaking changes
 - All URL's for uploaded images has changed (But we will implement a temporary fallback)
 - The docker compose stack has changed
 - New environment variables have been added for Minio
 - The upload API client and endpoints has changed

## Usage Examples

Upload an image:
```
POST /api/1.0/images
```

Upload a document (PDF):
```
POST /api/1.0/documents
```

Retrieve an image (original size):
```
GET /api/1.0/images/[image-id]
```

Retrieve a resized image:
```
GET /api/1.0/images/[image-id]?w=800
```

Retrieve a document:
```
GET /api/1.0/documents/[document-id]
```

This implementation provides a solid foundation for all our image and document handling needs while keeping security and performance at the forefront.


apps/frontend/src/lib/formio/providers/storage/beabee.ts
```diff
@@ -33,20 +33,22 @@ export default class BeabeeStorage {
...
+      const response = await client.upload.uploadFile(file);
+
+      // Direct access to the URL from the response
+      const imageUrl = response.url;
 
       return {
         storage: 'beabee',
         name,
-        url: url,
+        url: imageUrl,
```
JumpLink, line 51: this can also be a document, rename this

packages/core/src/services/DocumentService.ts
```diff
@@ -0,0 +1,329 @@
...
+      forcePathStyle: this.config.s3.forcePathStyle !== false
+    });
+  }
+
+  /**
+   * Sanitize filename to prevent path traversal attacks
+   * @param filename Original filename
+   * @returns Sanitized filename
+   */
+  private sanitizeFilename(filename: string): string {
```
JumpLink, line 57: Move this to utils

JumpLink:
I am currently trying (based on the main branch) to upload PDFs to test the migration script for PDFs as well. The frontend does not show any errors, but in the Web Inspector I get a message saying that PDFs are not supported:

grafik

docker-compose.yml
```diff
@@ -105,44 +102,28 @@ services:
...
     image: minio/mc:latest
     depends_on:
       minio:
         condition: service_healthy
-    entrypoint: >
-      /bin/sh -c "
-      mc alias set myminio ${MINIO_ENDPOINT:-http://minio:9000} ${MINIO_ROOT_USER:-minioadmin} ${MINIO_ROOT_PASSWORD:-minioadmin};
-      mc mb --ignore-existing myminio/${MINIO_BUCKET:-uploads};
-      mc policy set download myminio/${MINIO_BUCKET:-uploads};
+    entrypoint: |
```
wpf500, line 115: This sort of image configuration should be done as part of building the image. We used to use Docker Compose to override some build-level configurations before and it causes a lot of problems with migrations (especially for self-hosters who need to update their own configuration). It's better if we just provide a Docker image that is fully configured (apart from env vars of course!).

Also we should always lock image versions rather than using `latest` for security and to reduce the likelihood of unexplained errors occurring.

e.g.
apps/mc/Dockerfile
```Dockerfile
FROM minio/mc:<version>

COPY ./docker-entrypoint.sh /entrypoint.sh
ENTRYPOINT [ "/entrypoint.sh" ]
```
apps/bc/entrypoint.sh
```sh
#!/bin/sh
....
```

packages/core/src/services/DocumentService.ts
```diff
@@ -0,0 +1,331 @@
...
+        ? sanitizeFilename(originalFilename)
+        : undefined;
+
+      // Generate a unique ID for the document
+      const fileId = randomUUID();
+
+      // Use the original filename extension if available, otherwise default to ".pdf"
+      const extension = sanitizedFilename
+        ? getExtensionFromFilename(sanitizedFilename)
+        : ".pdf";
```
wpf500, line 116: Why default to PDF?

apps/frontend/src/store/generalContent.ts
```diff
@@ -12,6 +14,20 @@ export const initGeneralContent = client.content
     return content;
   });
 
+// Process the image URL - add prefix if needed
+function processImageUrl(url: unknown): string {
```
wpf500, line 18: Duplicate method with `AppImageUpload.vue`, can it be moved to a util or something?

apps/frontend/src/components/callout/CalloutSummary.vue
```diff
@@ -59,4 +60,10 @@ const props = defineProps<{
 }>();
 
 const calloutLink = computed(() => `/callouts/${props.callout.slug}`);
+const imageUrl = computed(() => {
+  if (props.callout.image) {
```
wpf500, line 64: Clearer as ternary?
```ts
props.callout.image ? `${env.apiUrl}/${props.callout.image}` : noImage;
```

wpf500 (changes requested):
Looks good so far. I don't fully understand everything and would like to do a bit more testing, but in general it looks good and it will be very nice to remove pictshare.

I have a few concerns to discuss:
* Support SVG upload comes with [security implications](https://www.securesystems.de/blog/svg-security-risks-not-just-a-scalable-graphic/). How are we guarding against those?
* Is upload rate limiting still in place? The old service had (different) rate limits for guests and users, this is important (especially for guests).
* Serving files through `/api/1.0`. It's probably fine for now but really static assets shouldn't be served by an API, they are much more cacheable and could be served by a different route that is more optimised for content delivery. Dynamic image generation does require some sort of middle man service I guess, but I would aim for this to be on a different thread to the API service in the medium term

apps/frontend/src/store/generalContent.ts
JumpLink: Good found, yes I will mobe it to utils

apps/frontend/src/components/callout/CalloutSummary.vue
JumpLink: Yes, thank you

packages/core/src/services/DocumentService.ts
JumpLink: Currently, only PDFs are allowed as document uploads, but it still makes sense not to limit the service to this, so I have adjusted it accordingly.

apps/backend/src/api/interceptors/ValidateResponseInterceptor.ts
```diff
@@ -13,13 +13,19 @@ import { validateOrReject } from "@api/utils";
 
 const log = mainLogger.child({ app: "validate-response-interceptor" });
 
+// TODO: Do not use this for all controllers and use `@UseInterceptor(ValidateResponseInterceptor)` instead?
```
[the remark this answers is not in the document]
JumpLink: Okay, I'll leave it as it is for now and remove the comment.

docker-compose.test.yml
```diff
@@ -52,40 +52,23 @@ services:
     environment:
       MINIO_ROOT_USER: ${MINIO_ROOT_USER:-minioadmin}
       MINIO_ROOT_PASSWORD: ${MINIO_ROOT_PASSWORD:-minioadmin}
+    env_file:
```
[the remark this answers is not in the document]
JumpLink: Is removed now

docker-compose.yml
JumpLink: Resolved with https://github.com/beabee-communityrm/monorepo/pull/175/commits/efb25c97990719bd3a40403e918c4fb7440355cb

JumpLink:
@wpf500 I'm now finished and have implemented everything that was mentioned in the review. All that's left to do is adjust the deploy script and the production docker compose stack, but that will be separate pull requests anyway.

PR #177, opened 2025-04-24
 fix: PDF uploads on local dev machine

JumpLink:
This fixes the PDF upload for local development

## Changes

- Updated the image version for the img_upload_app service in docker-compose.yml from v0.2.0 to v0.3.0.
- Modified the upload client to handle error responses more effectively, checking for a status in the response and throwing a ClientApiError if the file type is not allowed.


wpf500:
Closing as we will merge #175

PR #178, opened 2025-04-28
feat: suppress some response errors for guest submissions

wpf500:
This PR suppresses errors that occur when a guest response is converted into a contact response. This prevents the endpoint from being used to probe if a particular contact has responded to a callout. The guest will always see success, regardless of whether the response has really been stored or not.

The create callout response endpoint has therefore changed for non-logged in users (guests). When guests creates a response they are only given back the ID of the response, which could be randomly generated nonsense if the response wasn't created.

Logged in users:
```
POST /callout/slug/responses
Cookie: <as logged in user>

<response DTO>
```

Guests:
```
POST /callout/slug/responses

{"id": "xxx"}
```

### To do
- [x] Internationalise warning text

### DIFF
--- a/apps/backend/src/api/controllers/CalloutController.ts
+++ b/apps/backend/src/api/controllers/CalloutController.ts
@@ -33,6 +33,7 @@ import {
   CreateCalloutResponseDto,
   GetCalloutResponseDto,
   GetCalloutResponseMapDto,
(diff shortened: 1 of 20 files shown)


wpf500:
Merging for pre-release, still needs review

JumpLink (commented):
LGTM

PR #198, opened 2025-05-09
feat(locale): Add Weblate translation support

JumpLink:
This PR significantly enhances the locale package for better integration with Weblate, improves translation workflows, and introduces new configuration and tooling for language management.

## Changes

- **Comprehensive Documentation**
  - Added a detailed `README.md` for the locale package with usage instructions.
  - Included a translation guidelines section for contributors (referenced in Weblate).
  - Documented both translation workflows: Weblate (for non-developers) and direct JSON editing (for developers).
  - Linked to official Weblate documentation in multiple languages.
  - Added a Contributor Agreement section to clarify licensing terms for translators.
  - Ensured license information is properly documented (CC-BY-4.0) and added the license field to `package.json`.

- **Translation Workflow & Status**
  - Removed all references and scripts related to the old Google Sheets-based translation process.
  - Added a Weblate translation status widget for live progress visualization.
  - Updated the testing checklist and provided clear instructions for testing translations.
  - Migrated usage metadata from Google Sheets to Weblate.

- **Locale Configuration & Language Visibility**
  - Introduced a new `availableIn` property in the locale config to control where each language is visible (e.g., callouts, backend).
  - This allows disabling languages in specific contexts until they are fully translated or ready for use.

- **Tooling**
  - Added a new `@beabee/weblate-client` package to facilitate integration with the Weblate API.
    - This was initially used to migrate tags from Google Sheets to Weblate.
    - The migration script has since been removed, but the client package remains for potential future use or automation.

- **Other Improvements**
  - Added French as a new language for testing  (A few community translations have already been added directly.).
  - Updated and improved translations for several languages (German, Dutch, Portuguese, Russian, Italian, etc.).
  - Cleaned up and normalized translation files for consistency.
  - Added watch mode to rebuild locale files on change

## Benefits

- Better onboarding for new translators and contributors.
- Clear, standardized documentation and translation guidelines.
- Improved workflow for managing translations through Weblate.
- Enhanced transparency about translation status and progress.
- Fine-grained control over language availability in different parts of the platform.
- Clearer licensing terms for contributed translations.
- Foundation for future automation or integration with Weblate via the new client package.

## Translation Status

Translations are now managed via [Beabee Platform Project on Weblate](https://hosted.weblate.org/projects/beabee/platform/).

## Testing Checklist

- [x] Add test translations for existing strings in German and German (informal)
- [x] Verify Weblate automatically creates a PR with translation changes (see #199)
- [x] Test merging the automatic PR and check if translations are correctly applied
- [x] Add a new language through the Weblate interface
- [x] Verify the new language is properly integrated into our codebase
- [x] Test the full process with a non-developer contributor (thanks to @NathanBnm)
- [x] Migrate usage metadata from Google Sheet to Weblate (The labels `User`, `Admin`, `System` and `CNR` were transferred to Weblate via a script.)
- [x] Remove migration scripts after successful migration (Removed [here](https://github.com/beabee-communityrm/monorepo/pull/198/commits/9935dcb21f8638b15ab299a524b579d5ae23d33d))
- [x] Add the option to disable languages as long as they are not fully translated (via `availableIn` property)
- [x] Add watch mode for locale files

## TODO after merge

- [ ] Set Weblate base branch to main
- [x] Request free hosting at Weblate


JumpLink:
@NathanBnm We're just getting started with Weblate, and you've already contributed in French. Did you realise that?

Many thanks at this point :D

packages/locale/src/config.json
wpf500, line 1: Could this file just be removed in favour of just defining this directly in `config.ts`?

packages/locale/src/context.ts
wpf500, line 1: I think it might be better if this had a less generic name as I guess this is exported by the package? e.g. `LocaleContext` or something?

packages/locale/src/context.ts
```diff
@@ -0,0 +1,4 @@
+export enum Context {
+  Callout = "callout",
+  Backend = "backend",
```
wpf500, line 3: I find the name `Backend` a bit confusing. Does this mean the translations are available everywhere? In which case maybe `All` or something similar is clearer? For me backend makes me think of the API, which then makes me confused as the translations aren't used there

packages/locale/src/context.ts
JumpLink: I tried that first, but the problem is that the `Locale` type is generated from the properties in config.json. When I migrated config.json completely to config.js, I ended up with a circular dependency of the types.

packages/locale/src/config.json
wpf500: Ah if you move it into config.ts then Locale can just be defined from `config.ts` I think
```typescript
export type Locale = keyof config; // or something
```

packages/locale/src/context.ts
JumpLink: Theoretically, they are visible everywhere in the frontend, but we still have the admin distinction, which means that they are not visible everywhere. What is actually meant is the entire frontend (except for the callouts), but only if no other admin language is set. Would ‘frontend’ be a better name? Or simply `user` or `system`?

wpf500: Hmm yeah `system` sounds good for now, at least a bit clearer. I guess we might actually be able to merge `adminLocale` with `availableIn`, but no need in this PR

packages/locale/src/config.json
JumpLink: config would be of the type `Record<Locale, LocaleConfig>` which includes `Locale`. Or do you think we should just not specify a type so that it is determined automaticly? This is often relatively computationally intensive for the TypeScript Language Server. But defining the type `Locale` in this way too.. Just defining `Locale` manually? `type Locale = ‘en’ | ‘de’ | ..`

JumpLink:
Since French is now largely complete, I will activate it as an available language:
image

JumpLink:
Can now be tested on https://dev.beabee.io/

PR #216, opened 2025-05-27
fix(frontend): Gallery response panel not showing when clicking images

JumpLink:
## Problem
When I click on an image on https://abriss-atlas.ch/gallery/ a transparent mask appears, but nothing else.

grafik

## Solution

The problem was that the panel was looking for response number 0, but real responses have IDs like 123, 456. We fixed this by making `currentResponseNumber` a required prop and having the parent component pass the correct response number directly.

## Changes
- Made `currentResponseNumber` a required prop in `CalloutShowResponsePanel` instead of using complex fallback logic
- Updated gallery.vue to pass the selected response number explicitly to the panel component
- Updated image handling in gallery.vue and CalloutIntroPanel.vue to use `resolveImageUrl` utility for consistent image URL resolution

### DIFF
--- a/apps/frontend/src/components/pages/callouts/CalloutShowResponsePanel.vue
+++ b/apps/frontend/src/components/pages/callouts/CalloutShowResponsePanel.vue
@@ -36,7 +36,7 @@
 
 <script lang="ts" setup>
 import { faArrowLeft, faArrowRight } from '@fortawesome/free-solid-svg-icons';
-import { computed } from 'vue';
+import { computed, watch } from 'vue';
 
 import CalloutSidePanel from './CalloutSidePanel.vue';
 import {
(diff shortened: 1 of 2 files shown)


wpf500 (commented):
I think I would solve the response number problem by making `currentResponseNumber` a required attribute. Then you don't need any of the watch stuff or looping over responses inside the component, and it's the parent component's responsibility to pass in a valid response number, which feels valid.

In `CalloutShowResponsePanel`
```ts
const currentResponseNumber = defineModel<number>('currentResponseNumber', {required: true});
```

In `gallery.vue`
```vue
<CalloutShowResponsePanel
  :responses="[selectedResponse]"
  :current-response-number=selectedResponse.number
  ...
/>

JumpLink:
> I think I would solve the response number problem by making `currentResponseNumber` a required attribute. Then you don't need any of the watch stuff or looping over responses inside the component, and it's the parent component's responsibility to pass in a valid response number, which feels valid.
> 
> In `CalloutShowResponsePanel`
> 
> ```ts
> const currentResponseNumber = defineModel<number>('currentResponseNumber', {required: true});
> ```
> 
> In `gallery.vue`
> 
> ```vue
> <CalloutShowResponsePanel
>   :responses="[selectedResponse]"
>   :current-response-number=selectedResponse.number
>   ...
> />
> ```

Thanks, that's much better, I've changed it accordingly.

wpf500 (approved):

PR #217, opened 2025-05-30
refactor(workspace): Package clean-up

JumpLink:
## Changes
* Output directory unified to `dist`
* Missing watch and dev scripts added to packages
* Moved esbuild plugins and utils to a separate package
* Further package.json scripts standardised
* Version number adjusted to the actual current version number

I would like to do further cleanups such as further modularisation through new packages, but then in a separate PR so that they don't become too large.

 OP#573

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -7,6 +7,3 @@
 
 /apps/*/dist/
 /packages/*/dist/
-
-/apps/backend/dist/
-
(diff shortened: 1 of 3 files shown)


JumpLink:
> Would be good to check on dev as the built -> dist change could break something

Okay, I'll test it directly together with #218, this should also be tested on dev due to the Docker changes.

PR #220, opened 2025-06-03
fix(tests): Fix Docker Compose test setup hanging indefinitely

JumpLink:
### Problem
Docker Compose test stack hung during service startup because `app_router` was waiting for `app` and `frontend` services that weren't included in the test service list.

### Solution

**1. Override dependencies in test configuration**
- Updated `docker-compose.test.yml` to make `app_router` only depend on services actually started in tests (`api_app`, `webhook_app`)
- Removed dependencies on `app` and `frontend` services

**2. Optimized service startup**
- Temporarily excluded `app_router` from test service list since frontend services aren't needed for API tests
- Services started: `db`, `migration`, `api_app`, `webhook_app`, `stripe_cli`, `minio`

**3. Enhanced logging and error handling**
- Streamlined logging without timestamps for cleaner output
- Added try-catch blocks around setup/teardown processes
- Optional debug logging for API app container output

### Result
- ✅ Test setup no longer hangs
- ✅ Faster startup by avoiding unnecessary frontend services  
- ✅ Better debuggability with structured logging
- ✅ Flexible configuration for future extensions

### Testing
```bash
yarn test
DEBUG_LOGS=true yarn test  # with verbose logging
```


wpf500 (dismissed):

JumpLink:
I had to add the router again for the test to run.

wpf500 (approved):

PR #221, opened 2025-06-04
Refactor CLI commands to use dynamic imports for actions

JumpLink:
**Problem:**
CLI actions import `@beabee/core` which triggers expensive initializations during module loading, slowing down CLI startup time even when those initializations aren't needed for the specific command being executed.

**Solution:**
- Converted all static imports of action modules to dynamic `await import()` calls
- Made all command handlers `async` to support dynamic imports
- Actions are now loaded lazily only when the corresponding command is executed

**Benefits:**
- Faster CLI startup time
- Reduced initial bundle size
- Only necessary initializations run for the actual command being used
- Improved performance for users who only need specific CLI functionality

**Files changed:**
- `apps/backend-cli/src/commands/*.ts` - All command files updated to use dynamic imports

**Testing:**
All existing CLI commands should work exactly as before, but with improved startup performance.

### DIFF
--- a/apps/backend-cli/src/commands/api-key.ts
+++ b/apps/backend-cli/src/commands/api-key.ts
@@ -1,7 +1,3 @@
-import { createApiKey } from "../actions/api-key/create.js";
-import { deleteApiKey } from "../actions/api-key/delete.js";
-import { listApiKeys } from "../actions/api-key/list.js";
-
 import type { ArgumentsCamelCase, CommandModule } from "yargs";
 import type { CreateApiKeyArgs, DeleteApiKeyArgs } from "../types/index.js";
 
@@ -13,7 +9,10 @@ export const apiKeyCommand: CommandModule = {
       .command({
         command: "list",
         describe: "List all API keys",
-        handler: () => listApiKeys()
+        handler: async () => {
+          const { listApiKeys } = await import("../actions/api-key/list.js");
+          return listApiKeys();
+        }
       })
       .command({
         command: "create",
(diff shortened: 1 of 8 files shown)


wpf500 (commented):
Have you tested the cron jobs work? The cron environment is very specific, I don't know if `yarn` will be on the path

JumpLink:
@wpf500  You were right to be concerned - the cron jobs initially failed due to environment issues. The problem wasn't the PATH (yarn was available), but rather permission conflicts when using `su - node`, which tried to access `/root/.bash_profile` and `/root/.config/yarn`.

**Solution:** Changed to `su node -c 'export HOME=/home/node && ...'` which properly sets the HOME environment variable for the node user, allowing yarn to access its config directory correctly.

Confirmed working via Docker logs - the cron jobs now execute successfully every minute during testing. Will attach screenshot showing successful execution:

grafik

apps/backend/crontab
```diff
@@ -2,6 +2,6 @@
 PATH=/usr/local/bin:/usr/bin:/bin
 
 # Spread the jobs out over a few minutes to avoid overloading the server
-0 1 * * * su node -c 'cd /opt/apps/backend/ && sleep $((RANDOM \% 120)) && yarn backend-cli sync mailchimp' > /proc/1/fd/1 2>&1
-5 1 * * * su node -c 'cd /opt/apps/backend/ && sleep $((RANDOM \% 120)) && yarn backend-cli process gifts' > /proc/1/fd/1 2>&1
-10 1 * * * su node -c 'cd /opt/apps/backend/ && sleep $((RANDOM \% 120)) && node ./dist/tools/process-segments.js' > /proc/1/fd/1 2>&1
+0 1 * * * su node -c 'export HOME=/home/node && cd /opt/apps/backend/ && sleep $((RANDOM \% 120)) && yarn backend-cli sync mailchimp' > /proc/1/fd/1 2>&1
```
wpf500, line 5: Could this be above like the `PATH` definition rather than being per cronjob?

JumpLink: Good idea to running the cron jobs directly as the `node` user. I tried a few different approaches (environment variables in crontab, absolute paths, proper HOME setup) but kept hitting `exit status 2` errors. The cron environment for non-root users seems to have some quirks that are tricky to debug in containers.

PR #226, opened 2025-06-05
feat: add notice text field below payment button

wpf500:


apps/backend/src/api/dto/ContentDto.ts
```diff
@@ -212,6 +212,9 @@ export class GetContentPaymentDto implements ContentPaymentData {
 
   @IsNumber()
   taxRate!: number;
+
+  @IsString()
```
JumpLink, line 216: We should consider limiting the maximum length of such strings in general. E.g. like this:

```ts
@IsString()
@Length(0, 500)
noticeText!: string;
```

packages/locale/src/locales/de@informal.json
```diff
@@ -1325,7 +1325,9 @@
         "showNoContribution": "Registrierung ohne Zahlung erlauben",
         "suggestedAmounts": "Beiträge vorschlagen",
         "text": "In diesem Abschnitt kannst du dein Registrierungsformular bearbeiten, optimieren und ansehen.",
-        "title": "Registrierung"
+        "title": "Registrierung",
+        "noticeText": "",
+        "noticeTextHelp": ""
```
JumpLink, line 1330: A german translation could be:

```json
"noticeText": "Hinweistext unter dem Zahlungsbutton",
"noticeTextHelp": "Verwende dieses Feld, um einen Hinweis unter dem Zahlungsbutton hinzuzufügen, zum Beispiel zur Erklärung von Zahlungsperioden oder Bankeinzugsermächtigungen. Du kannst dieses Feld auch leer lassen."
```

JumpLink:
I merged main, reformatted the code and pushed everything during the review. Hope that's okay.

JumpLink (approved):
LGTM 👍 Just a general suggestion to limit the length of custom strings and suggestions for German translations.

I merged main, reformatted the code and pushed everything during the review. Hope that's okay.

apps/backend/src/api/dto/ContentDto.ts
wpf500: This is probably a good idea but I think there is already a backlog task for this sort of security/validation improvements so I will leave it in this PR

JumpLink (approved):

PR #228, opened 2025-06-06
fix(docker): Update health check endpoint in base Dockerfile

JumpLink:
- Changed the health check command in the base Dockerfile to point to the correct API health endpoint at `/api/1.0/health` instead of `/health` for improved monitoring.


JumpLink:
Works:
grafik

I will merge it without review, as it is a bug fix for a feature that has already been approved.

PR #274, opened 2025-07-16
feat: icons for mapping categories

RayMille:
## Feature Description
This is feature includes selection of color and icons for markers on the map (Feature 719) :

### Code changes:

- Adding new package `fontawesome` that enables searching fontawesome icons
- Adding a `AppIconPicker` component using the new package
- Some minor styling updates to `AppColorInput`

### UI changes:

1. Category Selection
Select a category

3. Style Selection
Select a Style
 
5. Icons in Map
  Colored icons in map

## Technical Updates

```
build: update NPM dependencies
```
New `fontawesome` package and dependencies have to be installed.

## Checklist before review

## Checklist before merging


packages/vue/src/components/form/AppIconPicker.vue
```diff
@@ -0,0 +1,90 @@
+<!--
+  # AppIconPicker
+  A icon picker that allows users to select icons from a predefined set. 
+-->
+
+<template>
+  <div style="margin-left: 8px">
```
wpf500, line 7: We should never have inline styles. Is this needed?

packages/vue/src/components/form/AppIconPicker.vue
```diff
@@ -0,0 +1,90 @@
...
+    <div class="grid grid-cols-[auto_1fr] items-center gap-2">
+      <font-awesome-icon
+        :icon="[modelValue.prefix, modelValue.name]"
+        style="font-size: 1.7rem"
```
wpf500, line 11: Same thing, either should be converted to using a Tailwind utility or removed

packages/vue/src/plugins/icons.ts
```diff
@@ -1,11 +1,11 @@
 import { type Library, library } from '@fortawesome/fontawesome-svg-core';
-import { faCircleNotch, faUser } from '@fortawesome/free-solid-svg-icons';
+import { faCircleNotch, faUser, fas } from '@fortawesome/free-solid-svg-icons';
 import { FontAwesomeIcon } from '@fortawesome/vue-fontawesome';
 import type { App } from 'vue';
 
 export const icons = {
   install(app: App): Library {
-    library.add(faUser, faCircleNotch);
+    library.add(faUser, faCircleNotch, fas);
```
wpf500, line 8: Why is this necessary?

apps/frontend/src/components/pages/admin/callouts/tabs/ResponseDisplayTab.vue
```diff
@@ -246,6 +246,79 @@
...
               </AppFormField>
+              <AppFormField>
+                <AppSelect
+                  v-model="localData.mapSchema.mapIconQuestion"
+                  :label="inputT('mapSchema.mapIconQuestions.label')"
+                  :items="mapIconQuestions"
+                />
+              </AppFormField>
+              <div v-if="localData.mapSchema.mapIconQuestion">
+                <h3 class="mb-2 font-semibold">
```
wpf500, line 257: Is this done elsewhere? Ideally we always use one of our typography components (e.g. `AppSubHeading`) rather than manually styling a `h3`

apps/frontend/src/components/pages/admin/callouts/tabs/ResponseDisplayTab.vue
```diff
@@ -406,6 +497,79 @@ const mapBounds = computed({
   },
 });
 
+// Define MapIconQuestionValue
+type MapIconQuestionValue = {
```
wpf500, line 501: Is this type necessary? We already have type definitions for most components

apps/frontend/src/components/pages/admin/callouts/tabs/ResponseDisplayTab.vue
```diff
@@ -246,6 +246,79 @@
...
+                  :label="inputT('mapSchema.mapIconQuestions.label')"
+                  :items="mapIconQuestions"
+                />
+              </AppFormField>
+              <div v-if="localData.mapSchema.mapIconQuestion">
+                <h3 class="mb-2 font-semibold">
+                  {{ inputT('mapSchema.mapIconStyling.label') }}
+                </h3>
+                <div
+                  v-for="(value, i) in getValues(mapIconQuestion)"
```
wpf500, line 261: I think it would be clearer to create a new component that encapsulates all the state for this functionality, e.g. (pseudo-code)
```vue
<MapIconPicker v-for="value in getValues(mapIconQuestion)" :value="value" />
```

apps/frontend/src/components/pages/admin/callouts/tabs/ResponseDisplayTab.vue
```diff
@@ -368,6 +448,11 @@ const formComponentItems = computed(() =>
       id: c.fullKey,
       label: c.label || c.fullKey,
       type: c.type,
+      multiple: c.multiple || false,
+      values:
+        c.type === 'select'
+          ? c.data.values
+          : (c.values as MapIconQuestionValue[]) || [],
```
wpf500, line 455: Should already be correctly typed?

packages/vue/src/components/form/AppColorInput.vue
```diff
@@ -23,6 +23,11 @@
     alpha-channel="hide"
     :color="modelValue"
     @color-change="changeColor($event.colors.hex)"
+    :style="
```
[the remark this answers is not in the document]
RayMille: I added the `font-family: inherit` because the external `ColorPicker` component was rendering a slightly different font for the word **Hue** than the application font. This way it would be aligned. If this was intended I can change it back. 

Without inherit:
Screenshot from 2025-07-21 12-30-20

With inherit:
Screenshot from 2025-07-21 12-29-56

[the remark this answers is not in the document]
RayMille: I see this is also done further down in the component but does not seem to take effect in my local environment. Let's discuss this when you are back. I am removing the `font-family: inherit` for now :)

packages/vue/src/plugins/icons.ts
RayMille: I was not able to find a better way to preload the icon set that I need in order for the icon search to work in the AppIconPicker...

wpf500 (changes requested):
I think that it might be possible to simplify the `MapIconPicker` model a bit, and a few other small changes, but otherwise looking great 👍

packages/vue/src/plugins/icons.ts
wpf500: Is it possible to do this just in the icon picker so it doesn't affect the page size for all users?

wpf500 (approved):

PR #281, opened 2025-07-24
feat: add search by email callout reviewer

RayMille:
This PR adds adding callout reviewers via email including notification messages.

Screenshot from 2025-07-30 16-17-36

Screenshot from 2025-07-30 16-17-16

## Checklist before merging

### DIFF
--- a/apps/frontend/src/components/callout/CalloutReviewerSearch.vue
+++ b/apps/frontend/src/components/callout/CalloutReviewerSearch.vue
@@ -3,7 +3,7 @@
     v-model="email"
     :placeholder="t('calloutReviewerManager.search.placeholder')"
     :disabled="props.disabled"
-    @update:model-value="handleSearch"
+    @update:model-value="handleSearch('email', $event)"
   />
   <AppNotification
     v-if="!emailExists && email.length > 0"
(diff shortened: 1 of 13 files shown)


apps/frontend/src/components/callout/CalloutReviewerSearch.vue
```diff
@@ -0,0 +1,92 @@
...
+const rules = {
+  v: {
+    email: () => emailExists.value,
+    notEmpty: not((value: string) => value === ''),
+  },
+};
+
+const validation = useVuelidate(rules, { v: emailExists });
+
+watch(validation, (newState) => {
```
wpf500, line 89: I don't think you should need this. Vuelidate does it's own event propagation, so if this component fails validation it will automatically fail in parent components too. Use still need to call `useVuelidate`, but everything else should handle itself.

We do use a `validation` event in the callout form builder, but actually this is a legacy from a previous implementation that should be cleaned up at some point

apps/frontend/src/components/callout/CalloutReviewerSearch.vue
```diff
@@ -0,0 +1,92 @@
+<template>
+  <AppSearchInput
+    v-model="email"
+    :placeholder="t('calloutReviewerManager.search.placeholder')"
+    :disabled="props.disabled"
+    @update:model-value="handleSearch('email', $event)"
+  />
+  <AppNotification
```
wpf500, line 8: I think it would be good to add a bit of top padding to the notifications, e.g. `pt-2` or `pt-4`.

wpf500 (changes requested):
Looks good, just a couple of quick changes requested. I haven't checked the validation one thoroughly, but I assume you were using it to invalidate the parent form?

wpf500 (approved):
:+1: :smile:

PR #282, opened 2025-07-29
feat: truncate url in map response side bar

RayMille:
This PR truncates the URL in the side bar of the map. This currently only applies in the map sidebar. 

Question: Do we want this in other places as well?

UI changes:
Screenshot from 2025-07-29 17-08-42

### DIFF
--- a/apps/frontend/src/components/form-renderer/form-renderer.css
+++ b/apps/frontend/src/components/form-renderer/form-renderer.css
@@ -336,8 +336,9 @@
   div[ref='value'] {
     @apply inline;
   }
-}
 
-.callout-form.is-simple .formio-component-url {
-  @apply inline-block max-w-full truncate;
+  /* Truncate URLs into one line and cut of with (...) */
+  .formio-component-url {
+    @apply inline-block max-w-full truncate;
+  }
 }


apps/frontend/src/components/form-renderer/form-renderer.css
```diff
@@ -337,3 +337,7 @@
     @apply inline;
   }
 }
+
+.callout-show-response-side-panel .formio-component-url {
```
wpf500, line 341: I don't think you need the extra class here, as the nested rule above (`.callout-form.is-simple .callout-form-renderer`) already targets the right place (the map response is just a response rendered in "simple" mode).

wpf500 (changes requested):

apps/frontend/src/components/form-renderer/form-renderer.css
```diff
@@ -337,3 +337,7 @@
     @apply inline;
   }
 }
+
+.callout-form.is-simple .formio-component-url {
```
wpf500, line 341: Sorry last comment! There is already a `.callout-form.is-simple` CSS block above (starts L307), I think it would be more logical to group this in there with other overrides (e.g. see the `.formio-component-file` rule)

wpf500, line 341: Sorry last comment, I could have been clearer last time. There is already a `.callout-form.is-simple` CSS block above (starts L307), I think it would be more logical to group this in there with other overrides (e.g. see the `.formio-component-file` rule)

wpf500 (changes requested):

apps/frontend/src/components/form-renderer/form-renderer.css
RayMille: Yes that makes sense, I overlooked that, sorry! Done :) ✅

PR #283, opened 2025-07-29
feat: fix conditional logic in response display

RayMille:
This pull request integrates the logic from the edit response mode into the display response mode. As a result, any slides that were skipped during the survey will not be shown in the responses. This change ensures a clear distinction between default values and the actual answers provided by users.


wpf500 (approved):
🎉

PR #284, opened 2025-07-30
fix: handle Mailchimp archived status and add fallback for new statuses

wpf500:
Archived was added as a status at some point. In general we should provide a fallback and send ourselves an alert when we detect an unknown status as this could indicate a change in Mailchimp's API.


RayMille (dismissed):
Looks good :)

PR #286, opened 2025-07-30
fix: pre-release testing issues

wpf500:
Changes:
* Fixed broken padding on join form
* Reverted handling of query params in `AppPaginatedTable` to fix pagination
* Updated locale template after building


RayMille (approved):
I tested it and it is working and it looks cleaner without emits 👍

PR #294, opened 2025-08-05
fix(docker): Fresh frontend service startup errors using default env vars

JumpLink:
### Problem
When setting up Beabee fresh, the frontend Docker service fails to start with `sed` errors like:
```
sed: Error: No such file or directory
sed: tracking: No such file or directory
sed: APPSIGNAL_KEY)#g: No such file or directory
```

This happens because the `docker-entrypoint.sh` script tries to replace environment variables with empty values, causing `sed` to interpret them as file paths.

### Solution
Added proper default values in `docker-compose.yml` for frontend environment variables:
- `CNR_MODE` and `EXPERIMENTAL_FEATURES`: Default to `false` (features disabled)
- `APPSIGNAL_KEY`, `CAPTCHAFOX_KEY`, `MAPTILER_KEY`: Default to `""` (empty string)

### Why this approach
- Keeps the existing `docker-entrypoint.sh` script unchanged
- Provides sensible defaults for optional features
- Ensures `sed` always receives valid string values instead of empty variables
- Reduces setup friction for new developers

This fix eliminates a common stumbling block during initial Beabee setup and makes the development environment more reliable out of the box.


RayMille:
> @RayMille Did you have similar problems with your first setup? With these changes, no more changes to the .env file should be necessary in order to be able to start the Beabee Docker Compose Stack, which should simplify the entry into development more again

Yes, I did. Sorry, I meant to report that, but it got lost among all the tasks on the first day. Really nice, that it's working now :)

wpf500 (approved):

Issue #172, opened 2025-04-15
Poor separation of concerns in NewsletterService


wpf500:
#160 introduced new tech debt to `ContactsService.updateContactProfile` to support merging newsletter groups. This was necessary because of some existing tech debt around the separation of data for the newsletter status/groups and the `NewsletterService`.

The `NewsletterService` currently has a dependency on the `ContactProfile` model which creates poor separation between the `ContactsService` and `NewsletterService`.  To fix this:

* Newsletter fields (`newsletterStatus` and `newsletterGroups`) should be put in a new model (e.g. `ContactNewsletter`)
* `ContactsService.updateContactProfile` should no longer handle any newsletter related fields
* A new method (e.g. `ContactsService.updateContactNewsletter`) should handle then instead.
* `NewsletterService` should remove all references or dependencies on `ContactProfile` or `contact.profile`
* Add a new route (`/contact/:id/newsletter`) to handle newsletter updates
  * Consider handling `groups` array updates in the same way as `tags` array updates are handled for consistency across the API, i.e. using the `+`/`-` syntax
    ```
    PATCH /contact/:id/newsletter
    {
      "status": "subscribed",
      "groups": ["+group1", "-group2"]
    }
    ```
* The route `/contact/:id/profile` should just handle profile updates

PR #320, opened 2025-08-21
fix(core): Trigger Double Opt-In process mailchimp

RayMille:
## Trigger Double Opt-in process

This pull request makes a change to the newsletter subscription flow in the `CalloutsService`. The newsletter status is now set to `Pending` instead of `Subscribed` to ensure that the Double Opt-In process is triggered for new subscribers.

- [ ] Done a self-review of my code

## Checklist before merging

### DIFF
--- a/packages/core/src/services/CalloutsService.ts
+++ b/packages/core/src/services/CalloutsService.ts
@@ -275,7 +275,7 @@ class CalloutsService {
       await ContactsService.updateContactProfile(
         contact,
         {
-          newsletterStatus: NewsletterStatus.Subscribed,
+          newsletterStatus: NewsletterStatus.Pending, // We need to send this to Pending first in order to trigger the Double OptIn.
           newsletterGroups: newsletter.groups,
         },
         { sync: true, mergeGroups: true }


wpf500 (commented):
That was simple!

Two questions:
1. I think with this implementation all users would get a double opt-in email, even if they were already logged in. Is this expected behaviour?
2. What happens to an already subscribed user? Do they go back to pending? Possibly we want to avoid this if so?

RayMille:
> That was simple!
> 
> Two questions:
> 
>     1. I think with this implementation all users would get a double opt-in email, even if they were already logged in. Is this expected behaviour?
> 
>     2. What happens to an already subscribed user? Do they go back to pending? Possibly we want to avoid this if so?

Hi Will,
I was just creating a similar question in Slack. I’m not sure what unintended behavior this could have. Thank you for adding these two points. Then I will test for these :)

packages/core/src/services/CalloutsService.ts
```diff
@@ -272,13 +273,19 @@ class CalloutsService {
 
     if (newsletter?.optIn) {
       log.info(`Opting contact ${contact.id} into newsletter`, { newsletter });
+      const profile = await ContactsService.loadContactProfile(contact.id);
```
wpf500, line 276: I think it might make more sense to put this in the `NewsletterService`, rather than exposing the `ContactProfile` to the `CalloutService`

My thinking is:
1. It feels like a nicer abstraction that the newsletter service has the logic that prevents you from putting a subscriber back into pending mode, i.e. this is a general rule to enforce rather than specifically in callouts
2. The contact profile is already loaded in the newsletter service, so this reduces the number of database queries
3.  The newsletter status really shouldn't be in the contact profile (this is longstanding tech debt that I somewhat documented here: #172), so it's better to not reinforce that pattern where possible.

wpf500 (changes requested):
Let's chat about the newsletter status checking logic in person, I have an idea of where it could go

wpf500 (approved):
🎉

PR #323, opened 2025-08-29
feat(frontend/form.io): Create custom provider that works with MapTiler

JumpLink:
This PR implements a custom MapTiler address provider for Form.io and restricts the form builder to use only our supported providers. The changes ensure that all form components use the correct providers and prevent errors from unsupported provider configurations.

### Key Changes

1. **MapTiler Address Provider Implementation**
   - Created `BaseAddressProvider` abstract class for extensible provider architecture
   - Implemented `MapTilerAddressProvider` with full geocoding support
   - Provider integrates with existing MapTiler client and supports international addresses
   - Results are transformed to Form.io-compatible format for seamless integration

2. **Provider Restriction & Default Configuration**
   - Restricted form builder to only enable MapTiler (address) and Beabee (storage) providers
   - Set MapTiler as default address provider for all address components
   - Set Beabee storage as default for all file upload components
   - Hidden provider/storage selection UI elements to prevent confusion

3. **Form Builder Configuration Updates**
   - Updated `formOpts` to automatically set correct providers for address and file components
   - Added CSS rules to hide provider/storage selection fields
   - Ensured all new forms use the correct provider configuration

4. **Database Migration for Existing Forms**
   - Created migration `1756466495968-UpdateCalloutFormSchemaComponents` to update existing callout forms
   - Automatically sets `storage: "beabee"` for all file components
   - Automatically sets `provider: "maptiler"` for all address components
   - Ensures backward compatibility for forms created before this change

5. **Type Safety & Architecture**
   - Added comprehensive TypeScript interfaces for provider options and results
   - Created extensible base classes for future provider implementations
   - Maintained backward compatibility with existing Form.io address component structure

6. **Dev CLI Fix**
   - Fixed `generate:index` script in frontend package.json (removed `start` command)

7. **Address Formatter & Testing**
   - Created `AddressFormatter` class for unified address data handling
   - Added comprehensive tests using Vitest framework
   - Supports pattern-based address formatting with fallback placeholders
   - Includes real-world test data from CERN, NASA, and Stonehenge
   - Transforms between MapTiler, Form.io, and unified address formats

8. **Router & Navigation Updates**
   - Updated route names in `CalloutMapHeader.vue` to use absolute paths
   - Updated gallery and map page routes to `/crowdnewsroom/[id]/gallery` and `/crowdnewsroom/[id]/map`
   - Updated typed router definitions to reflect new route structure

9. **Geocoding Integration**
   - Replaced legacy geocoding utilities with new `AddressFormatter`
   - Updated `reverseGeocode` to return `UnifiedAddress` instead of `GeocodeResult`
   - Integrated address formatting in map click handlers
   - Added support for address pattern formatting in callout responses

10. **Address Component Types**
    - Added `ADDRESS_COMPONENT_TYPE` enum for standardized address parsing
    - Added `UnifiedAddress` interface for consistent address data structure
    - Added `geocodeCountries` property to `CalloutMapSchema` for country restrictions

11. **Testing Framework**
    - Added Vitest as testing framework for frontend package
    - Created comprehensive test suite for `AddressFormatter`
    - Added test data from real geocoding results (CERN, NASA, Stonehenge)

### Technical Details

- **Address Provider**: Uses MapTiler's geocoding API with configurable language support and result filtering
- **Storage Provider**: Maintains existing Beabee storage implementation
- **Form Integration**: Automatically applies provider settings to prevent configuration errors
- **Error Prevention**: Hidden UI elements prevent users from selecting unsupported providers
- **Data Migration**: Existing forms are automatically updated to use correct providers
- **Address Formatting**: Supports placeholders like `{street_number} {address}, {postal_code} {municipality}` with fallback patterns
- **Testing**: Comprehensive test coverage using Vitest with real-world geocoding data

### Future Considerations

- Country restriction functionality for MapTiler provider (search for `geocodeCountries`)
- Additional provider support can be added by extending the base classes
- Provider selection UI can be re-enabled if multiple providers are needed

Based on the attached files, here's a brief description of the available placeholders:

### Available Address Placeholders

The address formatting supports the following placeholder types, derived from the `ADDRESS_COMPONENT_TYPE` enum:

### Basic Placeholders:
- `{street_number}` - House number (e.g. "90")
- `{address}` - Street name (e.g. "Hermannstraße") 
- `{postal_code}` - Postal code (e.g. "12051")
- `{country}` - Country (e.g. "Germany")

### Geographic Levels:
- `{municipality}` - Municipality/City (e.g. "Berlin")
- `{locality}` - District (e.g. "Mitte" in Berlin)
- `{region}` - Region/State (e.g. "Berlin")
- `{county}` - County (e.g. "Cuxhaven")
- `{subregion}` - Subregion (e.g. "Nordrhein" in NRW)

### Special Area Types:
- `{joint_municipality}` - Joint municipality
- `{municipal_district}` - Municipal district (e.g. "Mitte" in Berlin)
- `{neighbourhood}` - Neighbourhood (e.g. "Kreuzberg")
- `{place}` - Named place (e.g. "Brandenburger Tor")

### Usage:
- **Simple**: `{street_number} {address}` → "90 Hermannstraße"
- **With fallbacks**: `{municipality|county|region}` → uses first available component
- **Complete**: `{address} {street_number}, {postal_code} {municipality}` → "Hermannstraße 90, 12051 Berlin"

### Screenshots
[Video demonstration will be added here - showing the form builder with hidden provider options and address lookup functionality](https://github.com/user-attachments/assets/72c2f569-053c-4d3c-8e30-607c7244096e)

## Test
- [x] When a new upload form is created in a callout, provider selection should no longer be necessary and is now hidden
- [x] The same applies to addresses; autocomplete should work automatically
- [x] Address formatting should work as usual + the new one
- [x] Map should work as usual, search for place, set marker, etc

## Checklist before merging

- [x] Video demonstration added to PR description

## Ticket references:
- https://correctivdigital.openproject.com/wp/1058
- https://correctivdigital.openproject.com/wp/579
- https://correctivdigital.openproject.com/wp/1059
- https://correctivdigital.openproject.com/wp/1280


apps/frontend/src/components/form-builder/form-builder.interface.ts
```diff
@@ -174,6 +176,8 @@ export const formOpts = {
           icon: 'cloud-upload',
           schema: {
             type: 'file',
+            // We only support the beabee storage provider for file uploads
+            storage: 'beabee',
```
wpf500, line 180: Haha so simple 🤦

docs/frontend/maptiler-integration.md
wpf500, line 1: Is this helpful documentation? I think it would be very easy for the code to get out-of-date without any real benefit

packages/core/src/migrations/1756466495968-UpdateCalloutFormSchemaComponents.ts
wpf500, line 1: I think this should be a manual migration rather than an automated database one, i.e. maybe as a CLI command like the image migration.

This could break a callout if anyone has managed to define a custom provider. I think this is unlikely and it's great that we are removing this option going forwards, but I would still want to check this per-client before rolling the change out.

docs/frontend/maptiler-provider.md
wpf500, line 1: Is this documentation helpful? Shall we should discuss what sort of documentation we want in `docs/`? I would see this as being more useful as comments next to the relevant code.

apps/frontend/src/type/maptiler.ts
wpf500, line 1: Always so great how you improve the types as you go 🎉

wpf500 (changes requested):

wpf500:
These are some really great changes in general by the way, you always push the Form.IO stuff in a much better direction (like previously creating all the types for the different component answer schemas etc.).

Also cool to see that the address lookup "just works", but we'll need to consider the consequences of this a bit more before releasing.

packages/core/src/migrations/1756466495968-UpdateCalloutFormSchemaComponents.ts
wpf500: I don't think we should maintain backwards compatibility, but I do think we should know before we release which callouts could be affected so we can get in touch with the clients.

I wanted to avoid holding the release back by making the change opt-in, but actually now I think about it your automated migration is the best approach, and instead we can just temporarily pin the versions of the clients it could affect.

I agree we could use something that helps us easily run these sorts of reports, at the moment I use a script on the server to do it, but it's not helpful for others. For this release I suggest I use the old school method, and put this in our backlog together with the project of generally improving our infrastructure.

wpf500 (changes requested):
Some changes needed I think. We are really aiming for `CalloutResponseAnswerAddress` to be _the_ unified response address schema. I'm going to have a go at this

wpf500 (dismissed):

RayMille (dismissed):
To be honest, I find it a bit challenging to review this since there are many parts of the code I haven’t worked on yet. From my perspective, though, it looks good—I don’t see anything obviously problematic. I also ran it on my machine, and it works :) :tada:

wpf500:
This is ready, but because of the more complicated data migration we'll need to do a specific release for it with a few additional steps, so I'm leaving this unmerged for now.

apps/frontend/src/lib/index.ts
```diff
@@ -1,6 +1,9 @@
+export * from './address.formatter.test-data';
```
wpf500: This shouldn't be exporting test-data files. I guess this is auto generated and maybe needs an additional rule to exclude test-data files?

wpf500 (changes requested):
Last quick change before merging

apps/frontend/src/lib/index.ts
JumpLink: Thank you, good catch, fixed in https://github.com/beabee-communityrm/monorepo/pull/323/commits/9b634ad1610a55827147b265c556b4854b50d5dc

wpf500 (approved):

PR #330, opened 2025-09-15
fix(status-page): add i18n plugin for translations

wpf500:
This PR adds the `i18n` plugin to the status app so the `AppStatusPage` component can render. It looks like this was broken since #267. It also removes the props that were replaced by using internationalisation.


RayMille (approved):
Looks good to me!

PR #331, opened 2025-09-16
chore(callout-controller): increase body size limit from 100kb to 300kb

wpf500:


wpf500:
@RayMille I've merged this without approval as it's a live problem for the CNR team, but let me know what you think. I'm not sure if we should increase the body size across the whole API (currently anything that uses the `@Body` decorator has a 100kb limit) or not?

PR #332, opened 2025-09-16
fix(frontend): Can't filter CNR answers by bucket

RayMille:
This PR adds the possibility of setting the required field for the `AppSelect` for `enum` `FilterItems`. The default value is set to required, to prevent any changes to current filters.

(The implementation seems a little bit hacky, but I was not able to think of anything else at the moment. Happy for better suggestions.)

https://correctivdigital.openproject.com/projects/beabee/work_packages/1323/activity

## Checklist before merging

### DIFF
--- a/packages/vue/src/components/form/AppSelect.vue
+++ b/packages/vue/src/components/form/AppSelect.vue
@@ -59,14 +59,11 @@ export interface AppSelectProps<T extends string | number> {
 const emit = defineEmits<(e: 'update:modelValue', value: T) => void>();
 const props = defineProps<AppSelectProps<T>>();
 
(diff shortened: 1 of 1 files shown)


wpf500 (commented):
I would try a different route here as the field is required, it's just that the ID is empty string. I think the data validation in `AppSelect` is the underlying problem. It should validate if one of the options in `items` is selected, even if that option has an ID of empty string.

Ultimately the code causing the form not to validate is here:

https://github.com/beabee-communityrm/monorepo/blob/main/packages/vue/src/components/form/AppSelect.vue#L63-L65

Maybe there's a way to make this check if `v` is one of the item IDs, rather than just using `requiredIf`?

PR #348, opened 2025-10-08
fix(frontend/upload): Prevent Form Submission During File Uploads

JumpLink:
This PR implements upload tracking to prevent form submission while files are being uploaded. The solution uses a reactive counter in `BeabeeStorage` that disables form validation during active uploads.

I have not implemented a progress bar or spinner as this would require changes to form.io. 

[Internal Ticket](https://correctivdigital.openproject.com/projects/beabee/work_packages/768)

### DIFF
--- a/apps/frontend/src/components/form-renderer/FormRenderer.vue
+++ b/apps/frontend/src/components/form-renderer/FormRenderer.vue
@@ -28,7 +28,10 @@ import {
   faRemove,
   faTimesCircle,
 } from '@fortawesome/free-solid-svg-icons';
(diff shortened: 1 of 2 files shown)


JumpLink:
@wpf500 I've now added a `resetActiveUploadsCount` function to `BeabeeStorage` provider to call it in `FormRenderer.vue` `onBeforeMount` hook to ensures the upload counter is reset when navigating to new forms, preventing issues where unfinished uploads from previous pages block form submissions.

wpf500 (approved):

PR #350, opened 2025-10-10
fix(cnrMode): Force OnlyAnonymous access for all callouts in CNR mode

JumpLink:
In CrowdNewsroom (CNR) instances, callouts were still collecting contact information even though this feature should be disabled in CNR mode. The "collect contact information" setting is correctly hidden in the CNR UI, but existing callouts created before this fix were still configured to collect contact data.

## Root Cause
The contact information form is displayed when:
- User is not logged in AND
- Callout access is not `OnlyAnonymous`

However, the callout creation logic wasn't automatically setting access to `OnlyAnonymous` in CNR mode, allowing callouts to be created with other access levels that still enabled contact collection.

## Solution
Modified the `convertStepsToCallout` function in `apps/frontend/src/utils/callouts.ts` to automatically set `CalloutAccess.OnlyAnonymous` when `env.cnrMode` is active, regardless of the UI settings.

## Note
Existing callouts must be saved again for the change to take effect.

[Ticket](https://correctivdigital.openproject.com/projects/beabee/work_packages/1509)

### DIFF
--- a/apps/frontend/src/env.ts
+++ b/apps/frontend/src/env.ts
@@ -1,3 +1,8 @@
+const cnrModeRaw = '__cnrMode__' as unknown as boolean | string;
+const experimentalFeaturesRaw = '__experimentalFeatures__' as unknown as
+  | boolean
+  | string;
+
 export default {
   /**
    * Application base URL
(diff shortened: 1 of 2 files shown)


wpf500 (approved):
This is a good backstop, but I don't really understand how a user was able to make the callout anything other than `OnlyAnonymous` in the first case. In theory the logic in `SettingsTab.vue` should hide the options in `cnrMode`.

Could this have been some other strange edge case, e.g. we copied the callout from a non-CNR instance to a CNR one and didn't update the setting, or something?

PR #351, opened 2025-10-10
feat: Implement backend for callout response email notifications

JumpLink:
Implements backend functionality to send custom confirmation emails to callout participants who provide contact information. Builds on the frontend Email tab added in the previous PR.

This PR is part of several pull requests:
* #349
* #351 (You are here)
* #355 - merged into this one

Can be tested on [dev](https://dev.beabee.io/) (latest update 28 Oct 17:40).

### Database
- Add migration for `sendResponseEmail` flag on callouts
- Add `responseEmailSubject` and `responseEmailBody` fields to callout variants
- Fields support multi-language content via variant system

### Email System
- Create new `callout-response-answers` template in 6 languages (en, de, de@informal, it, nl, pt)
- Template includes proper structure with greeting, custom message, and footer
- Extend `EmailService.sendTemplateToContact()` to support custom subject override
- Merge fields supported: `*|FNAME|*`, `*|LNAME|*`, `*|EMAIL|*`, `*|MESSAGE|*`, `*|SUPPORTEMAIL|*`, `*|CALLOUTLINK|*`, `*|CALLOUTTITLE|*`

### Code Quality Improvements
- Extract merge field replacement logic into reusable utility functions (`replaceMergeFields()`, `expandNestedMergeFields()`)
- Add comprehensive documentation for all email template merge fields in `EmailService`
- Simplify `BaseProvider` by removing duplicate merge field replacement code
- Support nested merge fields: `MESSAGE` content can include merge fields that are automatically expanded (e.g., "Hello *|FNAME|*, thanks for responding to *|CALLOUTTITLE|*!")

### Business Logic
- Add `sendResponseEmail()` method in `CalloutsService`
- Email sent immediately after successful response submission
- Only sends when enabled and both subject/body are configured
- Logs warning if enabled but content missing

### API Updates
- Update DTOs and transformers for callout and variant models

Email is sent through existing template system ensuring consistent formatting and proper merge field replacement.

## Implementation Decision

**Current approach:** Store subject/body directly in `CalloutVariant` model and use generic `callout-response-answers` template as wrapper.

**Rationale:**
- Email content is callout-specific and not reusable across callouts
- Keeps email configuration directly with callout data (single source of truth)
- No additional Email entities to manage in database
- No lifecycle management needed (Email entities would need cleanup on callout deletion)

**Alternative approach (open for discussion):**
Use `Email` model + `EmailController` by creating an `Email` entity per callout with `responseEmailId` on callout.

Pros:
- Consistent with existing email management system (`membership-builder/email.vue`)
- Could use `client.email.update()` API in frontend
- Email editing UI already exists

Cons:
- Requires Email entity creation/deletion lifecycle management
- Additional database entries for callout-specific (non-reusable) content
- More complex: need to handle orphaned emails when callouts are deleted
- Overhead for storing essentially the same data in separate table

The current approach is simpler and better suited for callout-specific emails that aren't meant to be reused globally.

[Ticket](https://correctivdigital.openproject.com/projects/beabee/work_packages/1497)

### DIFF
--- a/packages/core/src/services/CalloutsService.ts
+++ b/packages/core/src/services/CalloutsService.ts
@@ -17,6 +17,7 @@ import { QueryDeepPartialEntity } from 'typeorm/query-builder/QueryPartialEntity
 import { v4 as uuidv4 } from 'uuid';
 
 import config from '#config/config';
+import { contactEmailTemplates } from '#data/email-templates';
 import { getRepository, runTransaction } from '#database';
 import {
   DuplicateId,
@@ -545,10 +546,13 @@ class CalloutsService {
       variant.responseEmailSubject,
       variant.responseEmailBody,
       {
-        mergeFields: {
-          CALLOUTTITLE: variant.title,
-          CALLOUTLINK: `${config.audience}/crowdnewsroom/${callout.slug}`,
-        },
+        mergeFields: contactEmailTemplates['callout-response-answers'](
+          contact,
+          {
+            calloutSlug: callout.slug,
+            calloutTitle: variant.title,
+          }
+        ),
       }
     );
   }
--- a/apps/legacy/src/apps/tools/apps/emails/app.ts
+++ b/apps/legacy/src/apps/tools/apps/emails/app.ts
@@ -85,6 +85,15 @@ const assignableSystemEmails: Partial<Record<EmailTemplateId, EmailType>> = {
     showContactFields: true,
     mergeFields: [],
   },
+  'callout-response-answers': {
+    name: 'Callout response confirmation',
+    showContactFields: true,
+    mergeFields: [
+      ['CALLOUTTITLE', 'Callout title'],
+      ['CALLOUTLINK', 'Callout link'],
+      ['SUPPORTEMAIL', 'Support email address'],
+    ],
+  },
   // General emails
   'confirm-email': {
     name: 'Confirm email',
--- a/packages/client/src/api/email.client.ts
+++ b/packages/client/src/api/email.client.ts
@@ -61,14 +61,15 @@ export class EmailClient extends BaseClient {
    * Supports all template types (general, contact, admin)
    *
    * @param type - The template type ('general', 'contact', or 'admin')
-   * @param templateId - The template ID (e.g., 'welcome', 'new-member')
-   * @param options - Preview options including merge fields and custom subject
+   * @param templateId - The template ID (e.g., 'callout-response-answers', 'welcome', 'new-member')
+   * @param options - Preview options including merge fields, custom subject, and locale
    * @returns The preview with merge fields replaced
    *
    * @example
    * // Preview a contact template
-   * await client.email.preview('contact', 'welcome', {
-   *   mergeFields: { REFCODE: 'AB123' }
+   * await client.email.preview('contact', 'callout-response-answers', {
+   *   mergeFields: { CALLOUTTITLE: 'My Callout' },
+   *   customSubject: 'Thank you for your response'
    * });
    *
    * @example
--- a/packages/core/src/data/email-templates.ts
+++ b/packages/core/src/data/email-templates.ts
@@ -272,6 +272,21 @@ export const contactEmailTemplates = {
     SPLINK: params.spLink,
   }),
   /**
+   * Email with callout response answers
+   *
+   * **Available Merge Fields:**
+   * - *|CALLOUTTITLE|* - Title of the callout
+   * - *|CALLOUTLINK|* - Link to the callout
+   */
+  'callout-response-answers': (
+    _: Contact,
+    params: { calloutSlug: string; calloutTitle: string }
+  ) => ({
+    CALLOUTTITLE: params.calloutTitle,
+    CALLOUTLINK: `${config.audience}/crowdnewsroom/${params.calloutSlug}`,
+    SUPPORTEMAIL: OptionsService.getText('support-email'),
+  }),
+  /**
    * Email when contribution didn't start
    */
   'contribution-didnt-start': (_: Contact) => ({


packages/core/src/utils/email.ts
wpf500: It would be nice to add some tests around these methods

packages/core/src/type/email.ts
```diff
@@ -26,6 +26,7 @@ export interface EmailAttachment {
 export interface EmailOptions {
   attachments?: EmailAttachment[];
   sendAt?: Date | undefined;
+  customSubject?: string;
```
wpf500: `EmailOptions` is used by a lot of methods in `EmailService` and its providers, but the new `customSubject` option is only used in template related methods.

I think a clearer implementation would be something like
```ts
export interface TemplateEmailOptions extends EmailOptions {
  customSubject?: string
}
```

And use this just for the methods where it makes sense.

wpf500 (changes requested):
Nice one.

One thing I noticed is that you've added per-variant email copy, but always send the default variant. This isn't a regression, we were previously only sending the default system language email, but given that we now have per-variant copy maybe it's worth putting a ticket in the backlog to make this actually be used? It would be a nice little improvement to make the system send user-localised emails rather than system-localised ones.

packages/core/src/services/CalloutsService.ts
```diff
@@ -501,9 +507,53 @@ class CalloutsService {
...
+
+    // Check if custom subject and body are configured
+    if (!variant.responseEmailSubject || !variant.responseEmailBody) {
+      log.warn(
+        `Callout ${callout.id} has sendResponseEmail enabled but no email content configured`
+      );
+      return;
+    }
+
+    // Send custom email using the callout-response-answers template
```
wpf500: These comments aren't explaining anything in this code, they are explaining the logic within EmailService. They would be more relevant as docs for the appropriate method within that service.

packages/core/src/services/EmailService.ts
```diff
@@ -310,12 +213,130 @@ class EmailService {
...
+      LNAME: contact.lastname,
+    };
+
+    // 3. Generate template-specific merge fields
+    let templateMergeFields: EmailMergeFields = {};
+    if (type === 'contact' && template in contactEmailTemplates) {
+      templateMergeFields = contactEmailTemplates[
+        template as ContactEmailTemplateId
+      ](contact, {} as any);
+    } else if (type === 'admin' && template in adminEmailTemplates) {
```
wpf500: Can remove the following two if statements, `templateMergeFields` is already an empty object

apps/backend/src/api/controllers/EmailController.ts
```diff
@@ -61,7 +59,94 @@ export class EmailController {
...
+      opts.locale = data.locale;
+    }
+    if (data.body) {
+      opts.body = data.body;
+    }
+
+    const preview = await EmailService.getTemplatePreview(
+      templateId,
+      contact,
+      opts
```
wpf500: Could just be `data`?

apps/backend/src/api/controllers/EmailController.ts
```diff
@@ -61,7 +59,94 @@ export class EmailController {
...
+  /**
+   * Preview a contact email template
+   * Available to all authenticated users
+   * Uses the current user's data for contact-specific fields
+   */
+  @Authorized()
+  @Post('/preview/contact/:templateId')
+  async previewContactEmail(
+    @CurrentUser({ required: true }) contact: Contact,
+    @Param('templateId') templateId: ContactEmailTemplateId,
```
wpf500: And here

apps/backend/src/api/controllers/EmailController.ts
```diff
@@ -61,7 +59,94 @@ export class EmailController {
...
+
+  /**
+   * Preview an admin email template
+   * Only available to administrators
+   */
+  @Authorized('admin')
+  @Post('/preview/admin/:templateId')
+  async previewAdminEmail(
+    @CurrentUser({ required: true }) contact: Contact,
+    @Param('templateId') templateId: AdminEmailTemplateId,
```
wpf500: And here

packages/core/src/services/CalloutsService.ts
```diff
@@ -501,9 +509,52 @@ class CalloutsService {
...
+      );
+      return;
+    }
+
+    await EmailService.sendCustomEmailToContact(
+      contact,
+      variant.responseEmailSubject,
+      variant.responseEmailBody,
+      {
+        mergeFields: contactEmailTemplates['callout-response-answers'](
```
wpf500: Switch out for simply inlining the method and can remove from `contactEmailTemplates`?

```ts
{
    CALLOUTTITLE: variant.title,
    CALLOUTLINK: `${config.audience}/crowdnewsroom/${callout.slug}`,
    SUPPORTEMAIL: OptionsService.getText('support-email'),
}
```

packages/core/src/services/EmailService.ts
```diff
@@ -336,19 +475,38 @@ class EmailService {
...
   }
 
+  /**
+   * Enrich merge fields with standard fields (SUPPORTEMAIL, ORGNAME)
+   * This ensures all templates have access to standard merge fields
+   *
+   * @param mergeFields Template-specific or custom merge fields
+   * @returns Merge fields enriched with standard fields
+   */
+  private enrichMergeFields(mergeFields: EmailMergeFields): EmailMergeFields {
```
wpf500: Feels a bit redundant compared to just doing the merge at the call site?

```ts
mergeFields: this.enrichMergeFields({
  ...getContactEmailMergeFields(contact),
  ...additionalMergeFields,
}),
// becomes
mergeFields: {
  ...getBaseEmailMergeFields(),
  ...getContactEmailMergeFields(contact),
  ...additionalMergeFields,
},
```

wpf500:
@JumpLink it felt to me redundant to just use the template code to generate the merge fields, given that it was no longer being sent via the time. But maybe that was wrong, I see you point that it now means there is no documentation about the email type and it's merge fields. Do you want to revert that change?

PR #339, opened 2025-09-24
refactor(492): test environment improvments

camillekndrs:
## Test Environment Improvements

This pull request introduces a comprehensive test data anonymisation and seeding workflow for the backend CLI, enabling the creation of anonymized database dumps (in JSON or SQL format) and the ability to seed the database from these dumps. The changes include new CLI commands, core anonymisation logic refactoring, and enhancements to model anonymisers to support more entities and fields.

**Test Data Workflow Enhancements**

* Added new CLI commands for anonymising and seeding the database, allowing export to JSON/SQL and import from JSON dumps. [[1]](diffhunk://#diff-f71bbaebd94b8f61e7a5a3bb12c5b7fb8f48bd050d932c7a47d3b8a8dfa06ffeL17-R65) [[2]](diffhunk://#diff-86839e2d2271e897737b528a468be40a176523c306d7ba4d904036dbfa965528R1-R64) [[3]](diffhunk://#diff-2d9485575edd1f4c0832c677421efbb7a2deab8852b967e12d00d0d9cdb9e38eR1-R17)
* Refactored anonymisation logic into `@beabee/core/tools/database/anonymisers/`, supporting both JSON and SQL dump formats, and implemented robust file handling for dump generation. [[1]](diffhunk://#diff-23fbaffad49b7fb0fc928ceeb3efccc6911e600e05b796d990fd3ea95ea1ce4aR2-R14) [[2]](diffhunk://#diff-23fbaffad49b7fb0fc928ceeb3efccc6911e600e05b796d990fd3ea95ea1ce4aR24-R112) [[3]](diffhunk://#diff-23fbaffad49b7fb0fc928ceeb3efccc6911e600e05b796d990fd3ea95ea1ce4aL42-R126) [[4]](diffhunk://#diff-23fbaffad49b7fb0fc928ceeb3efccc6911e600e05b796d990fd3ea95ea1ce4aL70-R156) [[5]](diffhunk://#diff-23fbaffad49b7fb0fc928ceeb3efccc6911e600e05b796d990fd3ea95ea1ce4aL143-R219) [[6]](diffhunk://#diff-23fbaffad49b7fb0fc928ceeb3efccc6911e600e05b796d990fd3ea95ea1ce4aL175-R340)

**Model Anonymisers Expansion**

[...]

PR #352, opened 2025-10-16
Refactor database tools and improve test environment setup

JumpLink:
This pull requests adds the existing anonymization scripts to the `backend-cli` and allows export of files as JSON format for usage in testing. It also adds additional documentation. JSON was chosen as a format in testing because it is easier to add additional test data manually. 

**Dependency and Project Structure Updates:**

* Adds the `csv-parse` package as a dependency for CSV parsing in the CLI.
* Updates project structure documentation to clarify the organization of actions, types, and utilities for the CLI.

**Developer workflow and code standards:**

- **Updated code and commit rules:**
  - Enforced a new guideline to always write code in the most human-readable and simple way.
  - Required that all AI-generated commits include "(generated by 🤖)" in the commit message for transparency.

### Testing
- [ ] All existing tests pass
- [ ] New database dump functionality verified
- [ ] Anonymization scripts tested with current schema

Related to : https://github.com/beabee-communityrm/monorepo/pull/339. 

(This PR was started without 🤖 and finished with 🤖, after a joined session with @JumpLink)


RayMille:
Hi @JumpLink, I can not assign you as a reviewer because the merge request was created by you. Let me see..

packages/core/src/tools/database/anonymisers/models.ts
```diff
@@ -323,10 +342,34 @@ export const contactTagAssignmentAnonymiser = createModelAnonymiser(
...
+  sentDate: () => new Date(),
+  recipients: () => [
+    {
+      email: chance.email({ domain: 'fake.beabee.io' }),
+      name: chance.name(),
+      id: uuidv4(),
+    },
+  ],
+  emailField: () => 'email', // Standard-Feldname
+  nameField: () => 'firstname', // Standard-Feldname
```
JumpLink: Should be in english or removed

.cursor/rules/git-commit-best-practices.mdc
```diff
@@ -27,6 +27,7 @@ Before committing:
 - **Explicit Approval Required**: Before executing any git commit, ALWAYS ask the developer for approval
 - **Show Changes First**: Present the planned commit message and changed files for review
 - **Wait for Confirmation**: Do not proceed with commit until explicit "yes" or "commit" instruction
+- **Transparency: For commits generated with AI using this workflow add (generated by 🤖) in the commit message
```
JumpLink: 👍

apps/backend-cli/package.json
```diff
@@ -17,6 +17,7 @@
     "@beabee/core": "workspace:^",
     "@inquirer/prompts": "^3.3.0",
     "chalk": "^5.4.1",
+    "csv-parse": "^5.6.0",
```
JumpLink: Can this be updated to the latest version `6.1.0`?

apps/backend-cli/src/actions/database/import.ts
```diff
@@ -0,0 +1,100 @@
...
+/**
+ * Import database from JSON or SQL dump file
+ *
+ * @param filePath Path to the dump file (optional, defaults to test data dump)
+ * @param type Import type: json or sql (auto-detected from extension if not specified)
+ * @param dryRun If true, only logs what would be done (JSON only)
+ */
+export const importDatabase = async (
+  filePath?: string,
+  type?: 'json' | 'sql',
```
JumpLink: Should be validated if `type` is `json`, `sql` or `undefined somewhere in this method

apps/backend-cli/src/actions/database/import.ts
```diff
@@ -0,0 +1,100 @@
...
+    const rl = readline.createInterface({
+      input: fileStream,
+      crlfDelay: Infinity,
+    });
+
+    let query = '';
+    for await (const line of rl) {
+      if (query) {
+        console.log('Running ' + query.substring(0, 100) + '...');
+        await manager.query(query, line !== '' ? JSON.parse(line) : undefined);
```
JumpLink: `JSON.parse` should be catched like this:

```ts
try {
  const params = line !== '' ? JSON.parse(line) : undefined;
  await manager.query(query, params);
} catch (error) {
  throw new Error(`Invalid JSON parameters on line ${lineNumber}: ${error.message}`);
}
```

apps/backend-cli/src/actions/database/import-callout-responses.ts
```diff
@@ -221,54 +228,67 @@ function createResponse(
...
+    ];
 
-  console.error(`Possible headers: ${headers.join(', ')}`);
+    console.error(`Possible headers: ${headers.join(', ')}`);
 
-  const rows = await loadRows(headers);
+    const rows = await loadRows(filePath, headers);
 
-  console.error(`Processing ${rows.length} rows`);
+    console.error(`Processing ${rows.length} rows`);
```
JumpLink: should be `console.log` instead of `console.error`

JumpLink (commented):
I can't review a PR created by myself, so here as comments.

Overall, it looks very good to me, I just have a few minor comments.

PS: You can also create a new PR of this to make clear, that this is your work and not mine if you want

packages/core/src/tools/database/anonymisers/anonymizer.ts
```diff
@@ -0,0 +1,248 @@
...
+ * Used when exporting models that don't contain sensitive data.
+ * Processes items in batches of 1000 for memory efficiency.
+ * Uses valueMap to remap foreign key references to anonymized models.
+ *
+ * @param model The model to export
+ * @param type Export format (json or sql)
+ * @param anonymiser Optional anonymiser to extract FK mappings from
+ * @param valueMap Map of old values to new values for FK remapping
+ */
+export async function exportModelAsIs<T extends ObjectLiteral>(
```
wpf500: I don't think this needs to be a separate method with heuristics for finding foreign keys, I think `anonymiseModel` above can be simply modified to map known values without running anonymisers

packages/core/src/tools/database/anonymisers/anonymizer.ts
```diff
@@ -0,0 +1,248 @@
...
+  for (const prop in objectMap) {
+    const propertyMap = objectMap[prop] as PropertyMap<unknown>;
+    const oldValue = item[prop];
+
+    if (oldValue && propertyMap) {
+      const valueKey = stringify(oldValue);
+
+      const newValue =
+        typeof propertyMap === 'function'
+          ? valueMap.get(valueKey) || propertyMap(oldValue)
```
wpf500: Something like the following would always try to map known values, but only conditionally anonymise

```ts
? valueMap.get(valueKey) || (anonymise ? propertyMap(oldValue) : oldValue)
```

apps/backend-cli/src/actions/database/export.ts
```diff
@@ -0,0 +1,283 @@
...
+
+  // ========================================
+  // Step 2: Export contact-related data
+  // ========================================
+
+  const contactAnonymisers = [
+    ...ALWAYS_ANONYMIZED_MODELS.filter(
+      (a) =>
+        a.model === Contact ||
+        (typeof a.model === 'function' &&
```
wpf500: What does this do?

wpf500 (changes requested):
The topological sorting stuff is fairly mad AI power 😅.

This looks generally really good, great to see all these tools out of the backend and that the anonymisation stuff is back :)

The biggest problem is the dependency on writing files to disk. It would be much better if the input was read from `stdin` and the output went to `stdout`. Reasons:

1. On the Hive it's much harder to write and read files in and out of containers
2. It's poor security to write to disk and soon all our containers will be migrated to read-only filesystems so this will break in any case
3. It would simplify a lot of things and remove all the file path detection stuff

PR #353, opened 2025-10-16
feat(ai): Port Cursor rules to Kilo Code rules / workflows

JumpLink:
Together with @RayMille, we tried out [Kilo Code](https://kilocode.ai/) as an open-source alternative to Cursor. As a first application test, I ported our existing Cursor rules to Kilo Code rules / workflows to be able to make use of our rules. This PR is the result of that.

[Ticket](https://correctivdigital.openproject.com/projects/beabee/work_packages/1544)

### DIFF
--- a/.kilocode/rules/frontend-code-structure.md
+++ b/.kilocode/rules/frontend-code-structure.md
@@ -1,0 +1,30 @@
+# Frontend Code Structure
+
+Whenever you write HTML or UI code, make sure to
+- Mobile-first design approach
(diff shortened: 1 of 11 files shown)


RayMille:
Review notes: I really like that Kilo Code is open-source and the flexibility in choice of models this allows us. 

I do not however love the fact that they added easy support and free integration of Grok per default. Maybe we can continue our research in some good alternatives :)

PR #477, opened 2026-02-16
feat(infra): use a message queue service for IPC

wpf500:
This is part of a series of PR to make beabee ready for Kubernetes.

To move to Kubernetes we need to remove the fixed Docker-style references to other services (e.g. `api_app`, `webhook_app`). This could be simply replaced with Kubernetes-style references, but to keep compatibility with Docker stacks (e.g. for self hosters), and because in general it's a better solution, this PR adds Redis as a message queue service for IPC.


packages/core/src/message-queue.ts
```diff
@@ -0,0 +1,54 @@
...
+interface Message {
+  event: string;
+  timestamp: string;
+}
+
+export const events = new EventEmitter();
+
+export async function connect(): Promise<void> {
+  await subClient.connect();
+  await subClient.subscribe('service:broadcast', handleMessage);
```
JumpLink: **No reconnect handling for subscriber**

The `redis` v5 client has auto-reconnect, but in subscriber mode subscriptions need to be re-established after reconnect. If the connection drops, the `service:broadcast` subscription will be lost silently. Not critical for dev, but important for production reliability.

packages/core/src/services/OptionsService.ts
```diff
@@ -107,7 +107,7 @@ class OptionsService {
 
     if (options.length) {
       await getRepository(Option).save(options);
-      await networkCommunicatorService.notifyAll('reload');
+      await broadcast('reload');
```
JumpLink: **Nit: self-notification**

The broadcasting service also receives its own message via Pub/Sub, so after `set()` the service triggers an extra `reload` on itself. The old `NetworkCommunicatorService` had the same behavior so this is not a regression — but worth either documenting as intentional or filtering out self-sent messages in a follow-up.

JumpLink (changes requested):
Good approach overall 👍

A few issues should be addressed before merging — see inline comments.

### Additional notes (not on diff lines)

**`serviceSecret` is now unused** (`packages/core/src/config/config.ts:166`)
`BEABEE_SERVICE_SECRET` / `config.serviceSecret` was only used by the removed `NetworkCommunicatorService` for JWT signing. Since Redis Pub/Sub doesn't need JWT auth, this config entry and env var can be removed — either here or as a follow-up.

wpf500:
Closing outdated PR, the diff can be used to implement this again more simply

PR #479, opened 2026-02-17
refactor(infra): move router app URLs to env vars

wpf500:
Second Kubernetes-related PR. This PR moves the hardcoded service names in the router into environment variables.

I've moved all the upstream servers into `upstream` declarations to make the `proxy_pass` DNS resolution static (i.e. at start time rather than per request). This means there is no need for a `resolver` declaration (which had a Docker specific IP address) and removes the overhead of DNS resolution (although in practice this was practically non-existent as the resolver was local).

Note: Originally I made the DNS resolution dynamic to account for the fact that a container's IP address could change on recreation, but the router wouldn't get restarted (and therefore would still have the old IP). I think this should be solved by the router being dependent on the upstreams, but this needs to be tested.

This PR will need a corresponding update in [beabee-deploy](https://github.com/beabee-communityrm/beabee-deploy) and [hive-deploy-stack](https://github.com/beabee-communityrm/hive-deploy-stack).

### DIFF
--- a/apps/router/nginx.conf
+++ b/apps/router/nginx.conf
@@ -1,3 +1,19 @@
+upstream api_app {
+    server ${API_APP_HOST};
+}
+
+upstream webhook_app {
+    server ${WEBHOOK_APP_HOST};
+}
+
+upstream legacy_app {
+    server ${LEGACY_APP_HOST};
+}
+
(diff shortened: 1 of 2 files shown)


wpf500:
Close in favour of #630

PR #497, opened 2026-03-06
feat(stripe-flow): support confirmation tokens in Stripe payment flows

wpf500:
This PR migrates from server-side Stripe setup intents to client-side payment intents with confirmation tokens. This enables proper on-session payments that are taken immediately, which is the correct approach for both subscriptions and one-time payments.

Frontend ticket: https://correctivdigital.openproject.com/wp/2303
Backend ticket: https://correctivdigital.openproject.com/wp/2304

It fixes the following problems:
- Showing 0€ during 3DS verification of one-time payments or subscriptions
- Having to attach then detach payment methods for one-time payments

It refactors some of the payment flow logic:
- Unified signup and payment flow language: start → advance → complete
- Separated our setup and advance params, and removed overly complicated generics for provider-specific params

The new flow requires the following breaking changes to signup and payment-related endpoints.
- `POST /signup`
- `POST /contact/:id/contribution`
- `PUT /contact/:id/payment-method`
- `POST /contact/:id/payment`
  - All changed how they take payment setup params (`completeUrl` and `paymentMethod` to `params`)
- `POST /signup/confirm-email?joinFlowId=<id>` -> `POST /signup/complete?id=<id>`
  - Renamed to unify language and remove `joinFlow` reference

**Note:** This ticket deliberately leaves a couple of to dos:
1. Handling requires_action from Stripe: will be addressed in the next PR
2. iDEAL payment handling: out-of-scope for now, will be addressed in a future PR
3. Manual-to-automatic conversions: out-of-scope for now, will be addressed in a future PR

### 1. Payment Flow Architecture
- ✅ Switched from setup intents to client-side confirmation tokens
- ✅ Implemented 3-step unified flow: start → advance → complete
- ✅ Payments now process immediately during session
- ✅ Proper use of payment intents for immediate payments

### 2. API Changes
- ✅ Replaced `PaymentFlowResultDto` with `PaymentFlowSetupResultDto`
- ✅ Added `PaymentFlowSetupParamsDto` and `PaymentFlowAdvanceParamsDto`
- ✅ Restructured signup flow with new `AdvanceSignupFlowDto` and `SignupCompleteDto`
- ✅ Updated all payment-related endpoints to use nested params structure

### 3. Technical Implementation
- ✅ StripeProvider now handles confirmation tokens
- ✅ StripePaymentForm creates tokens client-side
- ✅ Simplified PaymentFlowProvider implementations
- ✅ Added database migration for payment flow updates


apps/frontend/src/components/forms/StripePaymentForm.vue
```diff
@@ -176,15 +209,26 @@ onBeforeMount(async () => {
...
+      // if (clientSecret) {
+      //   const { error: actionError } = await stripe.handleNextAction({
+      //     clientSecret,
+      //   });
+      //   if (actionError) {
+      //     handleError(actionError);
+      //   } else {
+      //     emit('completed');
+      //   }
+      // }
```
JumpLink, line 231: Is this the `requires_action` TODO you want to implement in the next PR?
Maybe a `// TODO(requires_action): <please describe>` would be good here.

If you can sort that out straight away in the next PR, it can stay as it is; otherwise, I’d simply remove the code entirely or at least add a TODO note to it

Perhaps it would be a good idea if we agreed on a standard format for inline TODO comments in Beabee?

packages/core/src/providers/payment-flow/StripeFlowProvider.ts
```diff
@@ -1,113 +1,67 @@
...
-      completedPaymentFlow.mandateId
-    );
+  async getCompletedPaymentFlowData({
+    flow,
+  }: CompletedPaymentFlow): Promise<CompletedPaymentFlowData> {
+    if (!flow.params?.token) {
+      throw new NoPaymentMethod();
+    }
+
+    const token = await stripe.confirmationTokens.retrieve(flow.params.token);
```
JumpLink, line 58: Missing try/catch: As confirmation tokens are quite short-lived, it might be better to throw expired tokens as a domain-specific error (PaymentFailed or similar) so that the UX doesn’t simply run into a Stripe raw error.

JumpLink (approved):
LGTM! 🎉 The switch to confirmation tokens and the unified `start → advance → complete` vocabulary make the flow much cleaner, especially removing the provider-specific generics is a nice win. The migration is pleasantly conservative too: no destructive ops, existing data stays untouched. 👍

I put this review together with help from Claude Code, so please take it with that in mind, here are our findings, none of them blockers, happy to see them as follow-ups:

- **`StripePaymentForm.vue`**: The commented-out `handleNextAction` block (around line 222) is presumably the `requires_action` TODO from the PR description. Might be worth marking it as `// TODO(requires_action):` or removing it, so it's clear it's intentionally parked there.

- **`StripeFlowProvider.ts:58`**: `stripe.confirmationTokens.retrieve(...)` has no try/catch. Confirmation tokens are pretty short-lived, so it might be nicer to wrap expired/invalid tokens in a domain-specific error (`PaymentFailed` or similar) rather than letting the raw Stripe error bubble up to the user.

- **`CompletePaymentFlowDto.params`** is optional, but effectively required for Stripe. `StripeFlowProvider` currently catches that with `NoPaymentMethod` — works fine, but a check at the DTO boundary would be friendlier for API consumers.

- **`StripePaymentForm.vue`**: `firstName.value` / `lastName.value` are read both before AND after the `await createConfirmationToken`. Thanks to the loading guard on the button this is probably a non-issue in practice, but caching the values once at the start of the function would rule out any potential mismatch.

One quick sanity-check question: Stripe setup-intent flows that happen to be mid-3DS at deploy time would fail with `NoPaymentMethod` on `/complete`, right? If so — might be worth a note in the release plan so support has it on their radar.

Thanks for the great refactor! 🙌

apps/frontend/src/components/forms/StripePaymentForm.vue
wpf500: Will be fixed in next PR before merging this into main

wpf500:
> - **`CompletePaymentFlowDto.params`** is optional, but effectively required for Stripe. `StripeFlowProvider` currently catches that with `NoPaymentMethod` — works fine, but a check at the DTO boundary would be friendlier for API consumers.

I originally implemented this with provider-specific DTO requirements but this ended up adding a lot of complexity to the code (`PaymentFlow` had to become a generic, extra code to handle this, etc.) so I backed the changes out. I would prefer to keep this simple for now

PR #513, opened 2026-03-13
chore(dev): remove auto migrations on local setup

wpf500:
This PR removes automatic migrations on on local dev machines.

Currently it is impossible to run multiple downward migrations because `yarn typeorm:migrate` has a dependency on the `migration` service, which means every time it starts it runs all the migrations first!

In general I think we should manually apply migrations on our machines, similarly to how we need to sometimes do `yarn install` when dependencies change. The advantage of having them automatically run is outweighed by the inconvenience it causes. As well as currently not being able to revert migrations very easily, I also frequently encounter situations where a half written migration gets executed because Docker reloads or something, then I have to manually fix the database to represent the desired state.

For initial setup the docs already include `yarn setup` which runs the migrations in any case.

### DIFF
--- a/docker-compose.test.yml
+++ b/docker-compose.test.yml
@@ -68,14 +68,6 @@ services:
       - api_app
       - webhook_app
 
-  migration:
-    extends:
-      file: docker-compose.yml
-      service: migration
-    env_file:
-      - .env
-      - .env.test
-
   stripe_cli:
     extends:
       file: docker-compose.yml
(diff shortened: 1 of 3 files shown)


JumpLink (dismissed):

JumpLink (approved):

wpf500:
This PR unfortunately failed on the E2E tests. Now that they have been reworked I will create another PR for this.

PR #525, opened 2026-04-14
fix(error-code): catch more API errors appropriately

wpf500:
This PR improves the `isApiError` helper to catch all status codes by default (not just 400). This fixes a current bug where invalid email/password combinations were showing the generic error instead of the login failed error because only 400s were being checked.

grafik

I've checked all the call-sites for `isApiError` and cleaned a few for consistency. Changing the default parameter will have no impact on existing call-sites as all those that weren't specifying a status code array were already using a specific error code, so will continue to only catch that error.


RayMille (approved):
🎉

PR #529, opened 2026-04-15
chore(auth): improve logging for authentication actions

wpf500:
This PR adds better logging for some authentication actions like reset password, reset device, and login attempts. It also slightly improves the reset password endpoint to normalise the provided email address, as in other flows.


RayMille (approved):
🎉

PR #530, opened 2026-04-16
refactor(api-errors): introduce type-safe and shared errors across frontend and backend

wpf500:
This PR introduces strongly typed errors in the frontend and consolidates backend errors into a single, consistent system. The original reason for this was to be able to include structured metadata for the new `PaymentRequiresActionError` needed for the Stripe `requires_action` payment handling.

**Key improvements**

- Every API error type now has a unique code and a human-readable message.
- Errors can include structured metadata (for example, `TooManyRequestsError.retryAfter`), and the frontend now has the types needed to consume that data safely.
- All API-thrown errors now derive from our own error classes (instead of base routing-controller errors).
- This lays the groundwork for better Graylog logging, so we can more clearly see why errors were thrown.

**Why the diff is large**

A large portion of the diff comes from two expected changes:

1. Updating imports to use our error classes rather than `routing-controllers`.
2. Renaming backend error classes for consistency so they all end in Error (for example, `CantUpdateContribution` -> `CantUpdateContributionError`).

**Scope notes**

In a few places, I introduced new specific error types where we previously used generic BadRequestError. In general, we should continue moving away from BadRequestError where possible, but fully addressing that is outside the scope of this PR, and there are still cases where it remains appropriate.

**Follow-up opportunity**

One area still worth improving is the duplicated error class implementations between backend and client. Today, backend errors need to inherit from `HttpError` (from `routing-controllers`) so the router can handle them automatically, while that class is not available in the frontend. This could be improved in a future PR, but it felt too large for this one.


apps/backend/src/api/controllers/AuthController.ts
```diff
@@ -48,22 +47,14 @@ export class AuthController {
           // Forward HTTP errors
           if (err) {
             if (err instanceof HttpError) {
-              // Passport errors only have a `message` property, so we handle the message as code
-              if (err instanceof UnauthorizedError) {
```
wpf500, line 52: I removed this as it never did actually anything. To create an `UnauthorizedError` you have to provide a code, so `err.code ||= bar` is a non-op.

apps/backend/src/api/controllers/ContactController.ts
```diff
@@ -310,7 +310,10 @@ export class ContactController {
     @Params() { id }: { id: string }
   ): Promise<void> {
     if (id === 'me') {
-      await ContactMfaService.deleteSecure(target, data);
+      if (!data.token) {
```
wpf500, line 313: I moved the data validation from `ContactMfaService` to the controller to simplify the logic and controllers are the place to handle input validation where possible.

apps/backend/src/api/controllers/ImageController.ts
```diff
@@ -51,18 +54,19 @@ export class ImageController {
...
     }
 
     // Verify file type is allowed - multer handles size but we still check type
     if (!isSupportedImageType(req.file.mimetype)) {
-      throw new UnsupportedFileType({
-        message: `Unsupported image type ${req.file.mimetype}. Please upload a JPEG, PNG, WebP or AVIF image.`,
-      });
+      throw new UnsupportedFileTypeError(
+        req.file.mimetype,
+        ALLOWED_IMAGE_MIME_TYPES
```
wpf500, line 69: I switched to `ALLOWED_IMAGE_MIME_TYPES` instead of a hard-coded message

packages/client/src/api/upload-document.client.ts
```diff
@@ -28,42 +31,25 @@ export class UploadDocumentClient extends BaseClient {
...
     }
 
     const formData = new FormData();
     formData.append('file', file);
 
-    try {
-      const { data } = await this.fetch.post<UploadFileResponse>('', formData, {
-        dataType: 'multipart',
-      });
+    const { data } = await this.fetch.post<UploadFileResponse>('', formData, {
```
wpf500, line 49: Rethrowing the rate limit error isn't necessary, it already has a message from the API

packages/client/src/utils/fetch.ts
```diff
@@ -357,41 +358,12 @@ export class Fetch {
   protected handleErrorResponse<T>(
     result: FetchResponse<T>,
     response: Response
-  ): never {
-    if (result.data) {
-      if (isApiErrorResponse(result.data)) {
-        // Extract Retry-After header (seconds) if present
-        const retryAfterHeader = response.headers.get('Retry-After');
```
wpf500, line 364: There is no need to pass this for every error, it was only ever present in the case of a `TooManyRequestsError`. The error logic can now differentiate between different types of error data (the core purpose of this PR), so it just uses `TooManyRequestsError.retryAfter` where needed.

packages/core/src/errors/CantUpdateContributionError.ts
wpf500, line 1: All the files in `packages/src/errors` which have been added and deleted are because I took the opportunity to align the class names so they all end with `-Error`. In this case `CantUpdateContribution` -> `CantUpdateContributionError`

packages/core/src/utils/multer.ts
```diff
@@ -33,72 +32,53 @@ import {
...
-        message: `Unexpected file: ${error.field || 'unknown field'}: ${message}`,
-      });
+      return new UnsupportedFileTypeError(error.field || 'unknown');
 
     case 'LIMIT_FILE_COUNT':
-      return new BadRequestError({
-        message: `Too many files uploaded: ${message}`,
-        code: 'TOO_MANY_FILES',
-      });
+      return new FileUploadError(
```
wpf500, line 52: I introduced a new error `FileUploadError` which has the same status code (400) and moved the codes in this file to be in `subCode`. The frontend/client weren't using the sub-codes so nothing needs to change there.

packages/common/src/data/reset-security-flow-error-code.ts
```diff
@@ -2,7 +2,6 @@
  * Reset security flow error code enum.
  */
 export enum RESET_SECURITY_FLOW_ERROR_CODE {
-  NOT_FOUND = 'not-found',
```
wpf500, line 5: I switched to the more appropriate `NotFoundError` for this error state

packages/core/src/services/ContactsService.ts
```diff
@@ -562,31 +564,32 @@ class ContactsService {
...
-        code: RESET_SECURITY_FLOW_ERROR_CODE.NOT_FOUND,
-      });
+      throw new NotFoundError();
     }
 
     if (rpFlow.type !== RESET_SECURITY_FLOW_TYPE.PASSWORD) {
-      throw new BadRequestError({
-        code: RESET_SECURITY_FLOW_ERROR_CODE.WRONG_TYPE,
-      });
+      throw new ResetSecurityFlowError(
```
wpf500, line 575: I created the more clearly defined error `ResetSecurityFlowError` with the same status code as before (400), but with the codes moved to sub-codes

packages/common/src/data/login-codes.ts
```diff
@@ -2,6 +2,7 @@
  * Login codes
  */
 export enum LOGIN_CODES {
+  UNAUTHORIZED = 'unauthorized',
```
wpf500, line 5: Added a generic fallback code for `UnauthorizedError`

RayMille (dismissed):
Looks really cool 🎉

RayMille (approved):

PR #532, opened 2026-04-21
feat(email): make required email merge fields mandatory.

RayMille:
This PR ensures that merge tags in system-wide emails are required if they are mandatory for the application flow to work. For example, a reset password email should always contain a password reset link (*|RPLINK|*), as the application cannot function correctly without it.

It adds the according notification message as seen in screenshot below:
Screenshot from 2026-04-21 16-41-57

Related to: https://correctivdigital.openproject.com/projects/beabee/work_packages/2248/activity.


packages/locale/src/locales/en.json
```diff
@@ -1262,6 +1262,7 @@
     "notifications": {
       "created": "Email created successfully",
       "deleted": "Email deleted successfully",
+      "missingMergeTags": "This email is missing following required merge tags: {mergeTags}. Merge tags are placeholders for dynamic content (e.g., first name). Please add the missing tags to ensure your email works correctly.",
```
wpf500: Small typo: "This email is missing **the** following required merge tags"

wpf500 (changes requested):

packages/locale/src/locales/en.json
RayMille: Thank you. Fixed in: https://github.com/beabee-communityrm/monorepo/pull/532/commits/efce98c05f63ec05046fc3fefecc1a4be6aacafb.

wpf500 (approved):

PR #544, opened 2026-04-29
build: ESM migration foundation — TypeORM index generator and migration plan

JumpLink:
First in a stack of four ESM migration PRs. This one is foundation only — additive, no behavioral changes.

**Stack:**
- **#544** (this) — Foundation: TypeORM index generator + migration plan
- #545 — Atomic flip: `packages/core` + 4 backend apps to ESM
- #547 — `tsc`-only build, drop `@beabee/esbuild`, bump TypeScript 6

## What this adds

A `generate-class-index` command in `@beabee/dev-cli` that scans a directory and emits an explicit-import index file (`import { X } from './X.js'; export const xs = [X, ...]`). Replaces TypeORM's `__dirname + '/models/*.js'` glob pattern, which doesn't work under ESM. Generic enough to be reused for any "list of classes from a directory" use case.

Two name strategies:
- `filename` — class name matches file name (TypeORM entities)
- `parse` — class name extracted from `export class X` (TypeORM migrations)

`--check` mode flags drift in CI without writing.

Wires `generate:typeorm-index` into `@beabee/core` (consumed in #545).

Also adds `docs/esm-migration.md` — public-facing migration plan, removed in #546 once the migration is complete.

## Non-breaking

Additive only. `yarn check`, `yarn build`, `yarn test` continue to pass unchanged.


wpf500 (dismissed):
The migration plan looks solid.

My only question (and defo not a blocker) is what is the value of keeping `esbuild` in this tool chain? Is it possible/simple to just use `tsc` for everything? I think there are build performance gains but I would guess this is much reduced by the upcoming TS 7, and I think the `esbuild` stuff is just use for generating NPM packages?

JumpLink:
> The migration plan looks solid.
> 
> My only question (and defo not a blocker) is what is the value of keeping `esbuild` in this tool chain? Is it possible/simple to just use `tsc` for everything? I think there are build performance gains but I would guess this is much reduced by the upcoming TS 7, and I think the `esbuild` stuff is just use for generating NPM packages?

Good point; TSC should work fine without any issues. I’ve only kept esbuild in there because it’s significantly faster (it’s written in Go, just like TSC 7). But that’s still a good argument that it’ll soon be much faster in future too; I’ll see if I can just switch it over.

*Edit*
Now done with #547

PR #545, opened 2026-04-29
feat(esm): PR2 – flip packages/core and all backend apps to ESM

JumpLink:
Atomic ESM flip for `packages/core` and all four backend apps. Builds on #544.

**Stack:**
- #544 — Foundation: TypeORM index generator + migration plan
- **#545** (this) — Atomic flip: `packages/core` + 4 backend apps to ESM
- #547 — `tsc`-only build, drop `@beabee/esbuild`, bump TypeScript 6

## Why atomic

Core and its four consumers (backend, webhooks, legacy, backend-cli) all flip together. There's no clean intermediate state.

## Per-package changes

**`packages/core`**: `"type": "module"`, `.js` extensions on relative imports, TypeORM glob → auto-generated `entities.ts` / `migrations.ts`, `__dirname` → `fileURLToPath`, `require.resolve` → `createRequire`. Third-party CJS interop where Node's named-export detection can't see through to subpaths: gocardless enums via `gocardless-nodejs/types/Types.js`, jsonwebtoken via default + destructure, yaml-front-matter the same.

**`apps/backend` / `apps/webhooks`**: `"type": "module"`, `.js` extensions, slugify CJS interop. GoCardless `fetch-customer.js` / `fetch-data.js` rewritten as ESM.

**`apps/legacy`**: `"type": "module"`, `gulpfile.js` → `gulpfile.cjs` with explicit `--gulpfile ./gulpfile.cjs` (Gulp 4 doesn't auto-discover `.cjs`), JSON imports via `with { type: "json" }`, clean-deep / ajv / ajv-formats CJS interop, missed `module.exports = app` in `discourse/app.ts` flipped to `export default`.

**`apps/backend-cli`**: was already ESM. Confirmed clean.

## docker-compose

The previous `./packages:/opt/packages` mount on backend services exposed the full workspace — incompatible with the source-TS exports introduced by the ESM flip. Replaced with narrower per-package `dist/` mounts: `core/dist` (live-edited via `tsc --watch`), and `common/dist` + `client/dist` + `locale/dist` to keep the hot-reload workflow alive for those packages too. Devs run `yarn workspace @beabee/<pkg> watch` on the host alongside `docker compose up <service>` and edits surface into the running container.

## Why not tsx / ts-node / SWC?

Evaluated and rejected:
- **tsx / esbuild / Node-native TS**: cannot emit `emitDecoratorMetadata` (verified 2026-04). TypeORM, routing-controllers and class-validator all require it.
- **ts-node**: emits metadata correctly but is incompatible with TS 6 in non-`transpileOnly` mode and breaks on the codebase's many implicit type-only imports in `transpileOnly` mode (last release 2023).
- **SWC**: works in principle, but has known TypeORM forward-reference bugs (swc-node#705).

→ `tsc` remains the only viable build/run path for backend code.

## Generator-driven barrel files

`generate-index` now emits an `// Auto-generated by ...` header so the files don't drift. `@beabee/core/generate:index` switched from `--extension none` to `--extension js` so re-running reproduces the committed `.js` extensions. Re-generating all workspace barrels surfaced two missing exports in `apps/backend/src/api/transformers/index.ts` that had drifted out of sync.

## Verified

- `yarn check`, `yarn build`, `yarn test` (396+ tests across 6 workspaces) ✅
- Full Docker stack starts cleanly, migration runs, API health 200, auth flow + payment flow work


docker-compose.yml
wpf500, line 1: I'm not sure I understand the reason behind the changes here.

> The previous ./packages:/opt/packages mount on backend services exposed the full workspace — incompatible with packages that point exports to source TS. Replaced with a narrower ./packages/core/dist:/opt/packages/core/dist mount: backends live-edit core via tsc --watch, common/locale resolve to the image's compiled dist.

If possible we should keep hot-reloading for all packages.

~~Would something like this work?~~ (No! see below)

```yaml
x-backend-volumes: &backend-volumes
  - ./packages/core/dist:/opt/packages/core/dist
  - ./packages/common/dist:/opt/packages/common/dist
  - ./packages/locale/dist:/opt/packages/locale/dist

# ...

services:
  api_app:
    volumes:
      <<: *backend-volumes
      - ./apps/backend #...
```

Bah YAML doesn't support merging lists, so would need to be duplicated for each service

JumpLink: Good catch, addressed in [da7ef45](https://github.com/beabee-communityrm/monorepo/commit/da7ef45d11): added per-package `dist/` mounts for `common`, `client` and `locale` to all three backend services. Hot-reload workflow is now: `docker compose up <service>` plus `yarn workspace @beabee/<pkg> dev` on the host. Edits in `packages/common/src/**` rebuild dist on the host, the mount surfaces it into the container, and `node --watch` picks it up.

docker-compose.yml
```diff
@@ -44,7 +44,10 @@ services:
     env_file:
       - .env
     volumes:
-      - ./packages:/opt/packages
+      - ./packages/core/dist:/opt/packages/core/dist
+      - ./packages/common/dist:/opt/packages/common/dist
+      - ./packages/client/dist:/opt/packages/client/dist
```
wpf500: Client not needed, it's not a backend dependency I think?

JumpLink: Right — verified empirically:

- `@beabee/client` is not a declared dependency in any backend workspace (`apps/backend`, `apps/webhooks`, `apps/backend-cli`, `apps/legacy`, `packages/core`)
- No `import '@beabee/client'` anywhere in backend source

Removed the three `client/dist` mounts in [e60041622a](https://github.com/beabee-communityrm/monorepo/commit/e60041622a). Stack rebased.

wpf500 (approved):

docker-compose.yml
JumpLink: yes, thank you, fixed with https://github.com/beabee-communityrm/monorepo/pull/545/commits/e60041622a20cf6674b530d08a9707badb58669a

PR #547, opened 2026-04-30
feat(esm): PR4 – build with tsc only, drop @beabee/esbuild, bump TypeScript 6

JumpLink:
Final PR in the ESM migration stack. Builds on #546.

**Stack:**
- #544 — Foundation: TypeORM index generator + migration plan
- #545 — Atomic flip: `packages/core` + 4 backend apps to ESM
- **#547** (this) — `tsc`-only build, drop `@beabee/esbuild`, bump TypeScript 6

## What changes

Two commits:

1. **Switch published packages to `tsc`-only build** — `common`, `client`, `locale` build with a single `tsc` invocation that emits both `.js` and `.d.ts` into a flat `dist/`. The previous `tsc --emitDeclarationOnly` + `esbuild` two-step is gone.

2. **TypeScript 5.9.3 → 6.0.3** — preparation for TS 7. Required four small fixups:
   - `packages/fontawesome/tsconfig.json`: `moduleResolution: "node"` → `"bundler"` (`"node"` is `"node10"` in TS 6, deprecated)
   - `apps/legacy/src/core/app-loader.ts`: dynamic-import `assert:` → `with:` (assertion syntax replaced by import attributes)
   - `apps/dev-cli/tsconfig.json`: explicit `types: ["node"]` for stricter ambient resolution
   - `packages/vue/src/histoire.setup.ts`: `@ts-expect-error` on the CSS side-effect import

## Why tsc and not tsx/esbuild?

Same reason the backend stays on `tsc` (see #545): tsx and esbuild can't emit `emitDecoratorMetadata`. While the published packages don't use decorators themselves, switching only them to esbuild while keeping the backend on tsc would mean two build pipelines. One toolchain is simpler.

## Drops

- `@beabee/esbuild` package (~270 LOC of build infrastructure)
- `@gjsify/esbuild-plugin-transform-ext`, `esbuild` deps from common/client
- `dev-cli`'s `esbuild` command + action + types
- `packages/locale/esbuild.ts` (replaced by `packages/locale/scripts/build.mjs`)
- `COPY packages/esbuild` line from the Dockerfile

The locale-specific build helpers (JSON copy, fallback translations, template generation, normalisation) moved into `packages/locale/scripts/build.mjs` since locale was the only consumer.


wpf500 (approved):
🎉. Always happy to see our build simplified

PR #548, opened 2026-05-05
refactor(server): enable verbatimModuleSyntax across backend code

JumpLink:
Stacked on top of the ESM migration stack (#544 → #545 → #546 → #547).

## What this does

Adds `"verbatimModuleSyntax": true` to `packages/tsconfig/tsconfig.server.json`. From now on, every import that is type-only must be written as either `import type { X }` or `import { type X }` — the compiler rejects anything else.

## Why

`verbatimModuleSyntax` is now the enforcement mechanism for type-only imports across all backend code (`apps/backend`, `apps/webhooks`, `apps/backend-cli`, `apps/legacy`, `packages/core`). It's strictly stronger than an ESLint rule because:

- the type checker already runs in CI, so no extra tooling cost
- `tsc --build` enforces it, no escape via `eslint-disable`
- works without per-package ESLint config (which the backend apps don't have today)

This also future-proofs against any per-file ESM transpilation we might want to use later (tsx, ts-node `--transpileOnly`, SWC). Those break on implicit type-only imports — which `verbatimModuleSyntax` now makes impossible to introduce.

(Frontend already had this set via `tsconfig.vanilla.json`, which is why #546 added explicit type imports to common/client. This PR brings the backend in line.)

## How

1. Set `verbatimModuleSyntax: true` in `tsconfig.server.json` → 2,220 errors surface.
2. Run `eslint --fix` once with `@typescript-eslint/consistent-type-imports` against backend/core source — fixes ~390 files automatically.
3. Manually fix ~20 cases ESLint couldn't resolve without project-aware type info (mostly mixed type+value imports from the same module that need splitting).
4. One unrelated cleanup: dead `Auth` import in `apps/backend/src/api/utils/callouts.ts` removed (was unused, surfaced because the rule flagged it).

## Verified

- `yarn check` ✅
- `yarn build` ✅ (full topological build)
- `yarn test` ✅ (workspace unit tests; pre-existing e2e-api-tests failures unchanged — they require a running Docker stack)


JumpLink:
Closing — `verbatimModuleSyntax: true` is incompatible with `emitDecoratorMetadata: true` in this codebase.

**Why:** TypeScript with `emitDecoratorMetadata` historically keeps imports referenced by decorators even when they appear only in type positions. `verbatimModuleSyntax` overrides this — it elides anything written as `import type` regardless of decorator usage.

The ESLint autofix rule (`@typescript-eslint/consistent-type-imports`) cannot tell decorator-relevant types apart from regular type-only imports, so it converted DTOs and Param classes used in routing-controllers to `import type`. Result: at runtime the classes are `undefined`, class-validator throws *"unknown value was passed to the validate function"*, every route using those DTOs returns 400.

Reproduced live at `/api/1.0/content/general` — the GET handler validates `ContentParams { id }` via `@Params()`, the `ContentParams` class import was converted to `import type`, validation crashed. The error pattern would extend to every controller using DTOs/Params.

**Decision:** Stay with the current setup — backend keeps explicit and implicit type imports, no `verbatimModuleSyntax`. The future-proofing benefit (tsx, ts-node, SWC) is moot given those tools aren't viable here anyway (verified in the discussion on #545).

If we ever want explicit `import type` for clarity in non-decorator code, that would be a per-file/per-package opt-in — not a blanket tsconfig flip.

wpf500:
@JumpLink ah this is a real shame, would have been great to enable this. Thanks for trying. The experimental decorators thing is so embedded into our codebase with TypeORM and Routing Controllers (and supporting packages), feels like definitely something waiting to catch us down the line. Did you see any viable options for moving away from it? Or would it really require to use packages which aren't dependent?

PR #551, opened 2026-05-05
fix(payment): reset cancellation status on new recurring contribution

wpf500:
This PR fixes a bug where the user's cancellation status wasn't reset if they started a new recurring contribution. This bug was introduced in #496 when splitting out processing payment flow from update contribution flows.

Previous logic for starting new contributions was:
1. `PaymentService.updatePaymentMethod` -> sets the new payment method up
2. `ContactsService.updateContactContribution` -> creates the new contribution and clears any cancellation state

https://github.com/beabee-communityrm/monorepo/blob/856b9f4563a8da63086339d42a4cc36b0eb144c3/packages/core/src/services/PaymentFlowService.ts#L143-L149

The cancellation logic assumed that `ContactsService.updateContactContribution` would always be called and thus handle clearing the cancellation date, but in the new payment flow processing logic it is deliberately handled separately.

Ticket: https://correctivdigital.openproject.com/wp/2662

### DIFF
--- a/packages/core/src/services/PaymentService.ts
+++ b/packages/core/src/services/PaymentService.ts
@@ -140,9 +140,19 @@ class PaymentService {
       await getRepository(ContactContribution).save(contribution);
     }
 
-    return await new PaymentProviders[
-      newMethod as keyof typeof PaymentProviders
-    ](contribution).processPaymentFlow(flow as any); // TODO: improve type
+    const ret = await new PaymentProviders[newMethod](
+      contribution
+    ).processPaymentFlow(flow as any); // TODO: improve type
+
+    if (flow.form.action === 'start-contribution') {
+      // Clear any cancellation date as the contribution is being restarted
+      await getRepository(ContactContribution).update(
+        { contactId: contact.id },
+        { cancelledAt: null }
+      );
+    }
+
+    return ret;
   }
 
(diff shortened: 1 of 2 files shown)


packages/core/src/services/PaymentService.ts
```diff
@@ -140,9 +140,19 @@ class PaymentService {
...
+    const ret = await new PaymentProviders[newMethod](
       contribution
     ).processPaymentFlow(flow as any); // TODO: improve type
+
+    if (flow.form.action === 'start-contribution') {
+      // Clear any cancellation date as the contribution is being restarted
+      await getRepository(ContactContribution).update(
+        { contactId: contact.id },
+        { cancelledAt: null }
+      );
```
JumpLink: Would it make sense to create a helper method whose name is descriptive enough in itself that we can avoid having to add a comment?

JumpLink (dismissed):
That looks good; I just have one small suggestion

packages/core/src/services/PaymentService.ts
wpf500: What do you think of the change in 4dd84336ad6c0f39c12a27325f208125b243b97f?

JumpLink (approved):

PR #554, opened 2026-05-06
chore: upgrade yarn 4.12.0 → 4.14.1

JumpLink:

## Summary
Upgrades the project's pinned Yarn release from 4.12.0 to 4.14.1 (`yarn set version latest`).

`yarn 4.14` ships with `enableScripts: false` as the secure default — install scripts no longer run automatically as a hardening measure against supply-chain attacks. Rather than re-enabling them globally, this PR keeps that default and explicitly allow-lists the four packages that ship genuine native build steps via `dependenciesMeta`:

| Package | Why it needs install scripts |
|---|---|
| `@parcel/watcher` | Native FS watcher binary |
| `esbuild` | Native bundler binary |
| `sharp` | Native libvips bindings (image processing) |
| `vue-demi` | Install-time Vue 2 / Vue 3 entry-point switch |

Other packages whose only postinstall is a "thanks/donate" message or an optional native accelerator (`core-js`, `es5-ext`, `protobufjs`, `ssh2`, `cpu-features`) stay disabled and now produce a `YN0004` advisory instead of executing arbitrary install code.

The lockfile metadata version is bumped to v9 by the new release; no actual dependencies move.

## Test plan
- [x] `yarn install` clean from a fresh `node_modules` — only the four allow-listed packages run scripts; everything else is YN0004-skipped
- [x] `yarn check` ✅ (all workspaces)
- [x] `yarn build` ✅ (all workspaces)
- [x] `docker build --target builder -f packages/docker/base.dockerfile .` ✅
- [ ] CI green


wpf500 (approved):
Ah this is great!

wpf500 (approved):

PR #578, opened 2026-05-07
chore(deps): @tiptap/* 2 → 3

JumpLink:

## Summary
Bumps the TipTap suite from 2.27 to 3.22 across `packages/vue`. The bundled `@tiptap/extension-underline` is unchanged in our manifest — TipTap 3 still ships it as a separate package, so removing it would be a behaviour change unrelated to the bump.

### `Editor.commands.setContent` signature
TipTap 3 replaces the second `emitUpdate?: boolean` argument with a `SetContentOptions` object. Updated in `packages/vue/src/components/form/AppRichTextEditor.vue`.

### `Link` and `Underline` folded into `StarterKit`
TipTap 3 ships `Link` and `Underline` as part of `StarterKit`, so registering them as separate extensions triggers a `"Duplicate extension names found"` runtime warning. They're now configured via `StarterKit.configure({ link: { openOnClick: false } })` instead of separate `Link.configure(...)` / `Underline` entries.

### Transitive `prosemirror-*` deduplication
TipTap 3 ships its own `@tiptap/pm` wrapper that pulls in `prosemirror-model` `^1.25.4`, while several other consumers in the tree still request older caret ranges. Without help, `yarn install` produced **three** copies of `prosemirror-model` (1.24.1 / 1.25.1 / 1.25.4) and two each of `prosemirror-view` and `prosemirror-transform`. TipTap then mounts schemas from one copy and tries to manipulate fragments through another, which crashes at runtime as soon as the user hits Enter:

```
RangeError: Can not convert <> to a Fragment
(looks like multiple versions of prosemirror-model were loaded)
```

The four core ProseMirror packages are pinned via root `resolutions` so every consumer flattens onto the highest shared version. Verified clean with `yarn why prosemirror-model` (single 1.25.4 copy across all consumers).

## Verification
- [x] `yarn check` ✅
- [x] `yarn build` ✅
- [ ] CI green
- [ ] Manual smoke of the rich-text editor (callout description, email templates) — not covered by tests


package.json
```diff
@@ -28,7 +28,11 @@
   "resolutions": {
     "@types/express": "^4.17.21",
     "@types/express-serve-static-core": "^4.17.33",
-    "tar-fs": "^3.1.1"
+    "tar-fs": "^3.1.1",
+    "prosemirror-model": "^1.25.4",
```
wpf500: Why are these resolutions necessary?

packages/vue/src/components/form/AppRichTextEditor.vue
```diff
@@ -194,15 +192,15 @@ const { t } = useI18n();
 const editor = useEditor({
   content: props.modelValue as string,
   extensions: [
+    // TipTap 3 folded Link and Underline into StarterKit; configure them
```
wpf500: This comment is useful in the PR description but not relevant in the code itself. Can you merge it into the description?

wpf500 (changes requested):

package.json
JumpLink: These four ProseMirror packages need pinning because TipTap 3 ships `@tiptap/pm` against `prosemirror-model` `^1.25.4`, while several transitive consumers in the tree still request older caret ranges. Without the resolution, `yarn install` produced three copies of `prosemirror-model` (1.24.1 / 1.25.1 / 1.25.4) and two each of `prosemirror-view` and `prosemirror-transform`. TipTap then mounts schemas from one copy and tries to manipulate fragments from another, which crashes at runtime as soon as the user hits Enter:

```
RangeError: Can not convert <> to a Fragment
(looks like multiple versions of prosemirror-model were loaded)
```

I have moved the rationale into the PR description and verified the dedupe with `yarn why prosemirror-model` (single 1.25.4 copy across all consumers).

packages/vue/src/components/form/AppRichTextEditor.vue
JumpLink: Done — moved into the PR description (under "Breaking changes that hit our code") and dropped the inline comment in 9b54e9749b.

package.json
wpf500: Has this been reported upstream? This would be a good candidate for a PR in the problematic package instead of adding custom resolutions here.

wpf500 (changes requested):

package.json
JumpLink: Good question, turns out it's not really a TipTap bug. `@tiptap/pm@3.22.5` (and current `3.23.5`) declares permissive ranges like `prosemirror-model: ^1.24.1`, so their manifest is fine.

The duplicates came from yarn holding on to older lockfile pins from the TipTap 2 install — running `yarn dedupe` collapses everything to one version per package. I dropped the four `prosemirror-*` resolutions and pushed the deduped lockfile in 40c929347f. Verified one copy each of `prosemirror-model` / `-view` / `-transform` / `-state`, and `yarn check` + `yarn build` are green.

So nothing to file upstream, it was on our side.

wpf500 (dismissed):

wpf500:
@JumpLink it looks like some rich text inputs are unable to accept line breaks now (e.g. https://demo.beabee.io/admin/membership-builder)

There is the following error in the console:

grafik

JumpLink:
> @JumpLink it looks like some rich text inputs are unable to accept line breaks now (e.g. https://demo.beabee.io/admin/membership-builder)
> 
> There is the following error in the console:
> grafik

@wpf500 Fixed with https://github.com/beabee-communityrm/monorepo/pull/610

PR #579, opened 2026-05-07
chore(deps): Vite 6 → 8 cluster

JumpLink:

## Summary
Bumps the Vite ecosystem in `apps/frontend` and `packages/vue`:

| Package | Old | New |
|---|---|---|
| `vite` | ^6.4 | ^8.0 |
| `@vitejs/plugin-vue` | ^5.2 | ^6.0 |
| `@intlify/unplugin-vue-i18n` | ^6.0 | ^11.0 |

`vite-tsc` is moved separately (#577).

## Verification
- [x] `yarn check` ✅
- [x] `yarn build` ✅
- [ ] CI green
- [x] Manual smoke of `yarn dev` and the histoire dev server (`packages/vue`)

## Notes
Histoire (in `packages/vue/.histoire`) is still pinned to vite 6 internally — confirm the dev story server still boots after the bump. If not, follow-up will be needed once Histoire publishes a Vite 8 build.


wpf500 (changes requested):
The join form `v-model` changes aren't related to a dependency upgrade I think, and are related to an existing problem with modifying an model object's properties?

I would guess that neither `JoinFormStep1` nor `JoinFormStep2` ever really emit `update:model-value` so I think this fix is just hiding the compile time warning and not actually fixing the underlying problem. I would prefer either that the warning stays or the problem is properly fixed (not necessary in this PR unless the Vite upgrade really requires it).

JumpLink:
You're right. I reverted the three `v-model` edits locally on this branch, `yarn check` and `yarn build` for the frontend pass cleanly against Vite 8, the change isn't needed for the upgrade.

I'll drop it from this PR. The underlying issue (children mutating the model object instead of emitting `update:modelValue`) deserves its own PR.

wpf500 (approved):

PR #580, opened 2026-05-07
chore(deps): vue-router 4 → 5 + vue-accessible-color-picker 5 → 6

JumpLink:

## Summary
Migrates vue-router from 4 to 5 across `apps/frontend` and `packages/vue`, plus the linked `vue-accessible-color-picker` 5 → 6 bump.

### vue-router 5 surface that reached our code
- `useRoute('routeName')` — the typed-name overload was dropped; every call site drops the argument.
- `route.params.<x>` is now `string | string[]` instead of `string` for non-repeating segments. Each page casts at the source so downstream code keeps the same shape it had under v4.
- `route.name` is now `string | symbol` instead of just `string`. Touched in `apps/frontend/src/pages/admin/settings.vue` to guard before calling `startsWith`.

### vue-accessible-color-picker 6
- The `color-change` event detail changed from `{ colors: { hex } }` to `{ color: Color }` where `Color` is a `colorjs.io` instance. `AppColorInput.vue` now uses `$event.color.toString({ format: 'hex' })`.
- `colorjs.io` (~0.6.0-beta.3) added as explicit peer in `packages/vue`.

## Verification
- [x] `yarn build` ✅
- [x] `yarn test` ✅
- [x] Manual smoke of admin pages and the colour picker — type-checking is solid but the runtime behaviour of vue-router 5's matcher should be eyeballed (tab-state, deep links, query handling)

## Notes
`unplugin-vue-router` (0.19.2) still emits route names as string literals via the typed-router augmentation — those declarations are kept for documentation but no longer flow through to `useRoute`'s argument under v5.


JumpLink:
**Upstream fix for the `route.params.x as string` casts is on the way**

The casts are a direct consequence of `useRoute<Name>()` in vue-router 5.0.6 resolving `keyof RouteMap` to `never` — we diagnosed this in detail ([vuejs/router#2696](https://github.com/vuejs/router/discussions/2696)).

[vuejs/router#2700](https://github.com/vuejs/router/pull/2700) (merged 2026-05-06) fixes exactly that: the `TypesConfig.RouteNamedMap` augmentation moves out of the ESM `.d.mts` file into the ambient declaration generated via `routes.d.ts`, which always ends up in the tsconfig include. With that, typed routes work again in our setup too.

The fix is on `main` but **not yet released** — the latest tag is `5.0.6` (2026-04-22), and there are 34 unreleased commits since then.

### Plan
This PR waits for the next vue-router patch (presumably `5.0.7`). Once it's out:

1. Bump `vue-router` from ^5.0.6 to ^5.0.7
2. Mechanically remove the 22 `as string` casts in 17 pages
3. Enable the `useRoute<'routeName'>()` generic where needed instead

Until then the PR is functionally complete (all 17 affected pages with casts manually verified), just not cosmetically final.

JumpLink:
`5.0.7` is out now and works

packages/vue/package.json
```diff
@@ -44,14 +44,15 @@
     "@tiptap/vue-3": "^3.22.5",
     "@vuelidate/core": "^2.0.3",
     "color2k": "^2.0.3",
+    "colorjs.io": "~0.6.0-beta.3",
```
wpf500: `beta` versions shouldn't be in `package.json` unless absolutely necessary. In this case `~0.6.0-beta.3` resolves to `0.6.1` in any case (see `yarn.lock`), so can this be explicitly marked with `~0.6.1`?

wpf500 (changes requested):

packages/vue/package.json
JumpLink: done with https://github.com/beabee-communityrm/monorepo/pull/580/commits/6060b215391d8ac78be64242a0217d3159b17bb8

wpf500 (approved):

PR #581, opened 2026-05-07
chore(deps): maplibre-gl 5 + vue-maplibre-gl 5 + @maptiler/* 3

JumpLink:

## Summary
Bumps the map stack in `apps/frontend`:

| Package | Old | New |
|---|---|---|
| `maplibre-gl` | ^3.6 | ^5.24 |
| `vue-maplibre-gl` | ^3.1 | ^5.6 |
| `@maptiler/client` | ^2.2 | ^3.0 |
| `@maptiler/geocoding-control` | ^1.2 | ^3.0 |
| `geojson` | (transitive) | ^0.5 (added) |

### Code adjustments
- **`Map.loadImage`** is Promise-returning in maplibre-gl 5 — `apps/frontend/src/utils/images.ts` rewritten to `await mapInstance.loadImage(...).then(({ data }) => data)`.
- **`GeoJSONSource.getClusterExpansionZoom`** is also Promise-returning — `apps/frontend/src/pages/crowdnewsroom/[id]/map.vue` updated.
- **`@maptiler/geocoding-control@3`** dropped the DOM-style `addEventListener` and the standalone `style.css` export. The control now extends `Evented`, so the `pick` listener uses `.on('pick', handler)` and reads the new `event.feature`. The CSS is delivered through the Lit shadow DOM, so the previous CSS import is removed.
- **`vue-maplibre-gl@5`** internally imports `geojson` at runtime even though it declares it only as a devDependency — added as an explicit dep in `apps/frontend`.
- **Selection ring** — the previous `MglCircleLayer` against a dedicated `SELECTED_RESPONSE` GeoJSON source kept silently rendering nothing on maplibre-gl 5: the source held the feature, but no paint reached the canvas (queryRenderedFeatures returned `[]` even with forced `radius: 50`, `opacity: 1`, `color: red`). Replaced with a **raw maplibre `Marker`** with a custom HTML element (red border, transparent fill), driven by a single watch on `[selectedFeature, mapLoaded]`. Sidesteps the style/source pipeline entirely.
- **`queryRenderedFeatures` layer filter** — maplibre-gl 5 throws on unknown layer ids; v3 silently returned `[]`. The cluster-source `sourcedata` watcher can fire before the `unclustered-points` layer has mounted (it's guarded by `v-if="mapLoaded"` to wait for icons), so `findSelectedFeature` now filters its `layers` arg to the ones that actually exist.

## Verification
- [x] `yarn check` ✅
- [x] `yarn build` ✅
- [x] CI green
- [x] Manual smoke of the callout-map: cluster expansion ✅, geocoder pick ✅, custom icons ✅, selection ring around the clicked marker ✅

## Notes
The selection-ring stayed broken under several earlier attempts (always-mounted source with empty FeatureCollection, `markRaw`'d feature to dodge Vue reactivity, `circle-opacity: 0` instead of alpha-0 color, `v-if="mapLoaded"` ordering, explicit `moveLayer` after each selection). None of them produced a rendered ring, which is why this lands as a raw Marker rather than a fix to the source/layer setup.


JumpLink:
red circle is fixed now:
grafik

wpf500 (commented):
I'm not convinced by the replacement of `MglCircleLayer` with all this custom code. There must be an underlying reason why it has stopped working?

Unless there is a really good reason to upgrade this as a priority I would leave this change until there is time to look into the underlying reason.

wpf500:
Closing for now until this upgrade becomes urgent or we find a better solution for `MglCircleLayer`

PR #583, opened 2026-05-07
chore(deps): routing-controllers 0.10 → 0.11 + class-validator 0.14 → 0.15

JumpLink:

## Summary
Bumps the two pre-1.0 packages that gate each other (routing-controllers 0.11 declares `class-validator ~0.15` as a peer):

| Package | Old | New |
|---|---|---|
| `routing-controllers` | ^0.10.4 | ^0.11.3 |
| `class-validator` | ^0.14.1 | ^0.15.1 |

Workspaces affected: `apps/backend`, `apps/legacy`, `packages/core`.

## Verification
- [x] `yarn build` ✅
- [x] `yarn test` ✅ (incl. 17 e2e API suites and core unit tests)
- [x] CI green
- [ ] Manual smoke of validation error responses (e.g. POST /api/contacts with invalid payload) — class-validator 0.15 changes the order/format of some error messages, which is the most likely runtime delta

## Notes
This bump unblocks the Express 4 → 5 PR — routing-controllers 0.10 is incompatible with Express 5.


wpf500 (dismissed):

JumpLink:
@wpf500 Thank you, can this be merged?

wpf500:
@JumpLink has it been tested? I see the smoke tests are unticked

JumpLink:
> @JumpLink has it been tested? I see the smoke tests are unticked

Yes, I tested the other changes together with the others, but I didn't test the PR on my own. I haven’t actually carried out any specific controller tests, though. I’ve been doing things like creating new callouts in the frontend, etc., which means the controllers have been used via the API with this tasks

wpf500:
@JumpLink I see the following logs on `yarn install`

```➤ YN0060: │ class-validator is listed by your project with version 0.15.1 (p13ac03), which doesn't satisfy what routing-controllers requests (~0.14.1).
➤ YN0060: │ class-validator is listed by your project with version 0.15.1 (p234402), which doesn't satisfy what routing-controllers requests (~0.14.1).
➤ YN0060: │ class-validator is listed by your project with version 0.15.1 (p88f4b2), which doesn't satisfy what routing-controllers requests (~0.14.1).
```

I think the [package.json](https://github.com/typestack/routing-controllers/blob/v0.11.3/package.json) for `routing-controllers 0.11.3` declares `class-validator ^0.14.1` as a dependency, so a class validator upgrade isn't possible even with this version?

JumpLink:
@wpf500 Verified locally with 0.15.1 installed:

- It's a **peer** dependency, not a regular dep ([routing-controllers package.json](https://github.com/typestack/routing-controllers/blob/v0.11.3/package.json)) — Yarn warns but installs 0.15.1 fine, no resolution conflict.
- `class-validator@0.15.1` has exactly one breaking change vs 0.14.x: `@IsIBAN` now takes an options object instead of a positional arg ([typestack/class-validator#2618](https://github.com/typestack/class-validator/pull/2618)). We don't use `@IsIBAN` anywhere.
- routing-controllers' compiled output only calls `validateOrReject` at runtime (plus `ValidationError` / `ValidatorOptions` as types) — all unchanged in 0.15.
- `yarn workspace @beabee/core run test` passes (82/82), `yarn check` is clean.

So the peer range is just stale, not a real incompatibility — the warning is cosmetic. `typestack/routing-controllers` has no open issue tracking class-validator 0.15; worth filing a one-liner upstream asking them to widen `^0.14.1` → `^0.14.1 || ^0.15.0`. Until they do, we can silence the local warning via `packageExtensions` in `.yarnrc.yml` if it gets noisy.

wpf500:
I would keep the warning as a reminder as potentially future `class-validator` updates could cause problems, but given that the underlying problem is that `routing-controllers` is outdated (and seemingly unmaintained) your approach sounds good.

PR #584, opened 2026-05-07
chore(deps): Express 4 → 5 cluster (express, multer, helmet, body-parser)

JumpLink:

## Summary
Bumps the Express 4 stack across `apps/backend`, `apps/legacy`, `apps/webhooks`, `packages/core`:

| Package | Old | New |
|---|---|---|
| `express` | ^4.21 | ^5.2 |
| `@types/express` | ^4.17 | ^5.0 (root resolution) |
| `@types/express-serve-static-core` | ^4.17 | ^5.1 (root resolution) |
| `multer` + `@types/multer` | ^1 | ^2.1 / ^2.1 |
| `body-parser` | ^1.20 | ^2.2 |
| `helmet` | ^7.2 | ^8.1 |

### req.params shape
Express 5 makes `req.params.<x>` `string | string[]`. Each backend / legacy call site that consumed it as `string` now casts at the source — touched files:
- `apps/backend/src/api/decorators/TargetUser.ts`
- `apps/legacy/src/apps/login/app.ts`
- `apps/legacy/src/apps/members/apps/member/app.ts`
- `apps/legacy/src/apps/tools/apps/emails/app.ts`
- `apps/legacy/src/apps/tools/apps/exports/app.ts`

### Route syntax (path-to-regexp v8)
Express 5 ships path-to-regexp v8, which removes the regex-style `(?:...)` groups. `ContentController`'s `/:id(?:*)` (used to match multi-segment content IDs like `join/setup`) is migrated to the v8 wildcard `*id`. `ContentParams` now `@Transform`s the captured `string[]` segments back into the original `'join/setup'`-style ContentId before validation.

Without this fix the api_app crashed at startup with `TypeError: Unexpected ( at index 16: /1.0/content/:id(?:*)`, which is what surfaced as the 502 Bad Gateway responses in the e2e tests' previous run.

### Type resolutions
Root `resolutions` for `@types/express` / `@types/express-serve-static-core` moved to v5.

## Verification
- [x] `yarn build` ✅
- [x] `yarn test` ✅ — **all 17 e2e API suites pass (79 tests)** after the route fix
- [x] api_app boots clean in Docker, `/1.0/health` and the previously-broken `/1.0/content/general` return 200
- [ ] CI green


apps/backend/src/api/params/ContentParams.ts
```diff
@@ -1,8 +1,12 @@
 import { ContentId, contentIds } from '@beabee/beabee-common';
 
+import { Transform } from 'class-transformer';
 import { IsIn } from 'class-validator';
 
 export class ContentParams {
+  // Express 5 captures multi-segment wildcards as `string[]`; reassemble
+  // them to the original `'join/setup'`-style ContentId before validation.
+  @Transform(({ value }) => (Array.isArray(value) ? value.join('/') : value))
```
wpf500, line 9: Good solution. I think this is definitely an indicator that having `/` in content names was a bad idea (should have been `join-setup` instead?), but that's beyond the scope of this ticket to repair.

wpf500 (approved):

apps/backend/src/api/params/ContentParams.ts
JumpLink: I've create an issue for that: https://github.com/beabee-communityrm/monorepo/issues/605

PR #587, opened 2026-05-07
feat (payment): Add Google Pay and Apple Pay icons to payment options

aasthas9:
## feat: Change credit card payment icon to include Google and Apple Pay

1. Add property `showGoogleApplePay` to content table
2. Add new checkbox to toggle Apple and Google Pay icon visibility. When checked, the icons show up in the credit card payment button.

Note: The option does not change or get payment methods from Stripe. It assumes that the admin will enable this option only after enabling these methods in the Stripe account.

### Screenshots
- Notice text appears when checkbox is ticked
image
<br />

- Notice text disappears when checked is unticked
image

<br />
image

## Checklist before merging


apps/frontend/src/components/contribution/ContributionMethod.vue
```diff
@@ -76,6 +86,8 @@ export interface ContributionMethodProps {
   disabled: boolean;
   /** Title text for the payment method selection */
   title: string;
+  /** Whether Google and Apple Pay are enabled */
+  googleAndApplePayEnabled: boolean;
```
wpf500: I think it would be slightly cleaner to use the same name `showGoogleApplePay` for the property as in the join content. This helps people who read the code in the future more easily see the connection and improves "grepability" of the codebase.

apps/frontend/src/components/contribution/ContributionMethod.vue
```diff
@@ -41,7 +42,16 @@
           :aria-label="`Select ${methodLabels[method]} payment method`"
           @click="emit('update:modelValue', method)"
         >
-          <PaymentMethodIcon :method="method" /><span
+          <PaymentMethodIcon :method="method" />
```
wpf500: Moving the `<span>` onto a new line (instead of `/><span>`) has an impact here as Vue templates have subtle whitespace rules:

```html
<span>hello</span><span>world</span> -> "helloworld"

<span>hello</span>
<span>world</span> -> "hello world"

<span>hello</span><span
>
  world
</span> -> "helloworld"
```

The last one looks ugly but is the only way to have multiline syntax with no content spaces as the end and start tags have to touch (`><`).

This code here is a case where there was a deliberate attempt to control inline whitespace:
```vue
<PaymentMethodIcon :method="method" /><span
  class="m-2 inline">
  Card
</span>
```
Means there is no whitespace between the icon and text "Card", so the spacing is controlled by `m-2`. I think it would be better for consistency to preserve this behaviour so all spacing is consistent and just uses margins with no whitespace.

packages/locale/src/locales/en.json
```diff
@@ -1636,6 +1632,7 @@
         "noticeText": "Notice below payment button",
         "noticeTextHelp": "Use this field to add a notice below the payment button, for example explaining payment periods or bank authorizations. You can also leave this field blank.",
         "showAbsorbFee": "Ask user about absorbing the payment fee",
+        "showGoogleApplePay": "Enable Apple and Google Pay payments",
```
wpf500: I think this is confusing as this doesn't enable Apple/Google pay, just shows their icons.

- If it's unticked but they are enabled in Stripe, then users will see no icon but still have Apple/Google pay on the next screen
- If it's ticked but they are not enabled in Stripe, then users will see an icon but then have no option.

I think we should do two things to make this clearer:
1. Change the checkbox text to "Show Apple and Google Pay payment icons"
2. Add some help text below the icon using our existing `<AppInputHelp>` component "You must also enable Google and Apple Pay in Stripe, contact support for guidance"

Here's an example of that help component in action:

Image

wpf500 (changes requested):
Great first feature PR :). A few suggested small tweaks but overall really good :tada:

apps/frontend/src/components/contribution/ContributionMethod.vue
aasthas9: Same thing for the spacing between the new icons and the text, then?
**Not**
```
   <ApplePay class="mr-1" /><GooglePay class="h-4" />
</span>
<span>
```

but
```
   <ApplePay class="mr-1" /><GooglePay class="h-4" /> </span>
<span>
```

wpf500: More like:
```vue
   <ApplePay class="mr-1" /><GooglePay class="h-4" />
</span><span
>
```

It's very confusing because Vue _does_ remove whitespace between tags and inner content if it's on a new line (warning: plain HTML behaves differently!), but not newlines between inline tags!

```vue
<span>hello</span>
<span>world</span> -> "hello world"

<span>hello</span><span
>
  world
</span> -> "helloworld"
```

wpf500: I often just put it all on one long line then Prettier automatically formats it as necessary

```vue
<PaymentMethodIcon :method="method" /><span v-if...><ApplePay /><GooglePay /></span><span ...>{{ method }}</span>
```

Press save ;)

packages/locale/src/locales/en.json
aasthas9: Added help text and bound its visibility to `showGoogleApplePay`. See screenshots.

wpf500 (changes requested):
Thanks for the changes :). Last two things:

1. Our PR names use a conventional commit style format so the release changelogs are well formatted. Could you rename the PR? Something like "feat(payment): added Google and Apple Pay logo option".

2. For admin area strings we need translations for `de.json` (formal German) and `de@informal/json` (informal German). Unfortunately we don't have a particularly good system for this at the moment, you can choose your favourite translation tool.

wpf500 (approved):

PR #590, opened 2026-05-07
fix(frontend): hide map add button when maptiler key is missing

JumpLink:
## Summary

- Adding a new map response relies on MapTiler reverse-geocoding to resolve the clicked location to an address.
- When `BEABEE_MAPTILER_KEY` is not configured, the "add" button is now hidden in the map header, gallery header, and map overlays (embed button + mobile bottom bar), so users aren't offered an action that can't complete properly.

Complements #589, which fixes the crash when geocoding fails for an individual point.

## Test plan

- [ ] With `BEABEE_MAPTILER_KEY` set: add button visible in map header / gallery header / mobile bar / embed overlay as before.
- [ ] Without `BEABEE_MAPTILER_KEY`: no add button in any of the above; other header elements (variants, view link) unchanged.

### DIFF
--- a/apps/frontend/src/components/form-builder/form-builder.interface.ts
+++ b/apps/frontend/src/components/form-builder/form-builder.interface.ts
@@ -1,5 +1,6 @@
 import { type CalloutComponentSchema } from '@beabee/beabee-common';
 
+import env from '#env';
 import type { LocaleProp } from '#type';
 
 /**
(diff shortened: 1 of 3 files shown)


wpf500:
@JumpLink I also added a3f063d as I guess the address field also requires MapTiler?

wpf500 (approved):

PR #593, opened 2026-05-08
chore: automated release workflow with release-it

JumpLink:
## Summary

Adds automated release tooling so that publishing to npm, tagging, CHANGELOG generation, and GitHub release creation happen through a single `yarn release` command instead of being a manual, error-prone process.

> [!IMPORTANT]
> Once the merge is complete, releases must be created using `yarn release` in future, rather than directly via GitHub; the release will then be created on GitHub by the script

## Why

- Most `@beabee/*` packages have never been published. The one that has — `@beabee/beabee-common` — is many minor versions behind the repo (`0.21.1-alpha.2` on npm vs. `0.44.1` shipped on GitHub).
- Source manifests and shipped releases had drifted apart: every workspace `package.json` showed `0.28.1`, while the latest GitHub release tag is already `v0.44.1`. This commit set realigns them.
- Releases will become reproducible, the npm registry will stay current, and the GitHub release log doubles as a user-facing changelog.

## What changes for contributors

- **Releases are no longer drafted on github.com.** From now on, run `yarn release` locally on `main`. The GitHub release is created automatically and is marked as **latest** by default.
- For pre-releases, use `yarn release:rc` or `yarn release:beta`. These are marked as **pre-release** on GitHub and published to npm under the `next` dist-tag.
- `yarn release:dry` previews the next release without touching anything.

## Published packages

After merge, npm receives:
- `@beabee/beabee-common`
- `@beabee/client`

All other workspaces are marked `private: true` and stay internal. `@beabee/vue` and `@beabee/locale` are kept private for now — `@beabee/vue` is a candidate for deprecation in favour of Nuxt.UI, and `@beabee/locale` is its dependency. They can be flipped to public later if needed.

## Manual setup required before the first release

* [x] Add `NPM_TOKEN` (npm Automation token with publish rights on `@beabee`) as a repo secret.
* [ ] Confirm npm org access for `@beabee/client` — it has never been published before.
* [ ] Ensure GitHub Actions workflow permissions are set to "Read and write".
* [ ] Locally, export `GITHUB_TOKEN` (scope `repo`) in your shell before running `yarn release` — see the README. `gh auth token` works as a token source if you use the GitHub CLI: `export GITHUB_TOKEN=$(gh auth token)`.
* [ ] (Optional, recommended)* Configure npm Trusted Publishers for the two packages to drop `NPM_TOKEN` in favor of OIDC.

## How to test after merge (before the first real release)

The publish workflow includes a manual `workflow_dispatch` trigger with a `dry_run` input (default: `true`). GitHub only exposes this once the workflow is on the default branch, so it cannot be triggered from this PR — but it can be run immediately after merge without risk, since the dry run does not publish anything.

Once merged:

```bash
gh workflow run publish-npm.yml --ref main -f dry_run=true
gh run watch
```

The dry run executes the full pipeline:
- checkout, install, build
- `npm whoami` against the registry (validates `NPM_TOKEN`)
- `yarn pack` across the publishable workspaces (validates each tarball builds)

If the run is green, everything is ready for `yarn release`. If it fails, nothing has been published yet, so any fix lands in a follow-up PR.

[Ticket](https://correctivdigital.openproject.com/projects/beabee/work_packages/2200)


package.json
```diff
@@ -13,8 +14,12 @@
     "test": "yarn workspaces foreach -v -W run test",
     "check": "yarn workspaces foreach -v -W -p run check",
     "generate:index": "yarn workspaces foreach -v -W run generate:index",
-    "publish:next": "yarn workspaces foreach -v -W --no-private npm publish --tolerate-republish --tag next --access public",
-    "publish:latest": "yarn workspaces foreach -v -W --no-private npm publish --tolerate-republish --tag latest --access public",
+    "publish:next": "yarn workspaces foreach -v -W --no-private npm publish --tolerate-republish --tag next --access public --provenance",
+    "publish:latest": "yarn workspaces foreach -v -W --no-private npm publish --tolerate-republish --tag latest --access public --provenance",
+    "release": "GITHUB_TOKEN=${GITHUB_TOKEN:-$(gh auth token)} release-it",
```
wpf500: I would prefer not to have a baked in dependency on `gh`, can this be made optional here or implemented differently?

wpf500 (changes requested):
Thanks, looks good!

One thing: I think `@beabee/vue` and `@beabee/locales` should also be private, do you have any particular reason for publishing them?

JumpLink:
> Thanks, looks good!
> 
> One thing: I think `@beabee/vue` and `@beabee/locales` should also be private, do you have any particular reason for publishing them?

My thinking behind this was that it might be useful to use `@beabee/vue` to build, for example, a custom contribution widget. 

_In future, this could perhaps also be used to implement ‘custom extensions’ for Beabee, should this feature ever become available. In my view, extensions could be displayed in an iFrame and the extension developers would be responsible for hosting themselves, this is similar to how it’s done in Shopify._

`@beabee/locale` is a dependency of `@beabee/vue` and would therefore also need to be available.

package.json
JumpLink: > I would prefer not to have a baked in dependency on `gh`, can this be made optional here or implemented differently?

is now optional: https://github.com/beabee-communityrm/monorepo/pull/593/commits/dbe5e3e55f266b86e5fe66e123cabd3552a86a11

wpf500:
> > Thanks, looks good!
> > One thing: I think `@beabee/vue` and `@beabee/locales` should also be private, do you have any particular reason for publishing them?
> 
> My thinking behind this was that it might be useful to use `@beabee/vue` to build, for example, a custom contribution widget.
> 
> _In future, this could perhaps also be used to implement ‘custom extensions’ for Beabee, should this feature ever become available. In my view, extensions could be displayed in an iFrame and the extension developers would be responsible for hosting themselves, this is similar to how it’s done in Shopify._
> 
> `@beabee/locale` is a dependency of `@beabee/vue` and would therefore also need to be available.

Given that we are talking about deprecating `@beabee/vue` in favour of Nuxt.UI (or similar), I would prefer that this package remains private until either we decide to support custom extensions or we switch away and can just delete it.

JumpLink:
> Given that we are talking about deprecating `@beabee/vue` in favour of Nuxt.UI (or similar), I would prefer that this package remains private until either we decide to support custom extensions or we switch away and can just delete it.

Done with https://github.com/beabee-communityrm/monorepo/pull/593/commits/6bc9b1fe2d8862b1199a7ceeed5d6fa179dd77e5,

wpf500 (dismissed):

Issue #567, opened 2026-05-07
deps: FontAwesome 6 → 7


JumpLink:
Follow-up from the dependency-update sweep.

### Packages
| Package | Current | Target |
|---|---|---|
| `@fortawesome/fontawesome-svg-core` | ^6.7 | ^7.x |
| `@fortawesome/free-brands-svg-icons` | ^6.7 | ^7.x |
| `@fortawesome/free-regular-svg-icons` | ^6.7 | ^7.x |
| `@fortawesome/free-solid-svg-icons` | ^6.7 | ^7.x |

Workspaces affected: `packages/vue`, `packages/fontawesome`, `apps/frontend`, `apps/legacy`.

### Risk / scope
- FontAwesome 7 renamed several icons and dropped a few legacy aliases. The repo currently bundles its own icon manifest in `packages/fontawesome` — that script needs to be re-run and the visual diff sanity-checked.
- The `@fortawesome/vue-fontawesome` peer requirement may force a bump there too (currently ^3.2).

JumpLink:
Closing — converted to a PR per follow-up plan.

PR #595, opened 2026-05-11
fix(frontend): move loading icon back to center

wpf500:
This PR fixes the loading icon no longer being in the center. This bug was probably introduced in #567, but I haven't checked this.

Before:
image

After:
image


libby-correctiv (approved):
Fantastic

PR #597, opened 2026-05-12
test: Set up browser testing

aasthas9:
# Set up browser testing

## Browser tests in Playwright
1. Added test for join flow with one-time payment: navigate to homepage, make one-time contribution, check contribution email, set a password, login and look for contribution details
2. Test only runs for Firefox currently

## Test environment and Github workflow
1. Added two test data dumps - default (contains a few users and payments) and custom (non-anonymised emails, one-time payments enabled in content)
2. Stripe secrets injected via Github secrets into the custom data dump before import
3. Created JSON test data: Used *during* tests (eg. to enter user info, set passwords, etc)
4. If test fails, the database is exported and uploaded as artifacts, along with logs from Docker and Playwright

## Misc
1. Minor fixes in anonymiser - added missing fields that were not being anonymised, leading to FK errors on import

## Checklist before merging


apps/browser-tests/src/tests/joinFlow.spec.ts
```diff
@@ -0,0 +1,222 @@
+import { test, expect } from "@playwright/test";
+import userInfo from "../fixtures/user-info.json";
+
+let firstName: string = userInfo.firstName;
```
wpf500: I think most of these can be `const` instead of `let` as they never change?

apps/browser-tests/src/tests/joinFlow.spec.ts
```diff
@@ -0,0 +1,222 @@
...
+      "Contribute button enabled",
+    ).toBeEnabled(); // If the form was filled correctly, button should be enabled
+
+    // Proceed to payment page
+    await page.getByRole("button", { name: /contribute/i }).click();
+
+    await page.locator('input[name="firstName"]').fill(firstName);
+    await page.locator('input[name="lastName"]').fill(lastName);
+
+    await page
```
wpf500: Possibly a bit cleaner to store the content frame in a variable and reuse?

```ts
const stripeFrame = page.locator(...).first().contentFrame()

await stripeFrame.locator("#payment-numberInput").fill("...");
await stripeFrame.locator("#...")
```

apps/e2e-api-tests/vitest.config.ts
wpf500, line 1: Why was this file deleted?

docker-compose.test.yml
```diff
@@ -28,6 +28,8 @@ services:
     env_file:
       - .env
       - .env.test
+    volumes:
+      - ./packages/core/src/migrations:/opt/packages/core/src/migrations
```
wpf500: Why is this here?

apps/frontend/docker-entrypoint.sh
```diff
@@ -1,20 +1,13 @@
...
-      -e s#__appUrl__#$APP_BASE_URL#g \
-      -e s#__apiUrl__#$API_BASE_URL#g \
-      -e s#__revision__#$APP_REVISION#g \
-      -e s#__version__#$APP_VERSION#g \
-      -e s#__appsignalKey__#$APPSIGNAL_KEY#g \
-      -e s#__captchafoxKey__#$CAPTCHAFOX_KEY#g \
-      -e s#__maptilerKey__#$MAPTILER_KEY#g \
-      -e s#__cnrMode__#$CNR_MODE#g \
-    {} +
+  find /usr/share/nginx/html -type f -name '*.js' -exec sed -i -e "s|__appUrl__|$APP_BASE_URL|g" -e "s|__apiUrl__|$API_BASE_URL|g" -e "s|__revision__|$APP_REVISION|g" -e "s|__version__|$APP_VERSION|g" -e "s|__appsignalKey__|$APPSIGNAL_KEY|g" -e "s|__captchafoxKey__|$CAPTCHAFOX_KEY|g" -e "s|__maptilerKey__|$MAPTILER_KEY|g" -e "s|__cnrMode__|$CNR_MODE|g" {} +
```
wpf500: I'd be really interested to know why this fixes the problem as it looks pretty ugly and reduces readability quite a lot. If it's what is necessary then it's fine, but maybe we could have a quick look together to understand what's going on?

.github/workflows/browser-tests.yml
```diff
@@ -10,9 +10,78 @@ on:
...
+        id: docker_build
+        run: docker compose -f docker-compose.test.yml build
+
+      - name: Start Docker stack
+        id: docker_start
+        run: docker compose -f docker-compose.test.yml up -d
+
+      - name: Import test data
+        run: |
+          cat ./apps/browser-tests/src/fixtures/test-db-default.sql | docker compose -f docker-compose.test.yml exec -T api_app yarn backend-cli database import
```
wpf500: I don't know for sure if GitHub Actions will accept it, but if it does I would suggest this as slightly more readable code:

```
cat ./apps/browser-tests/src/fixtures/test-db-default.sql |
  docker compose -f docker-compose.test.yml exec -T api_app yarn backend-cli database import

cat ./apps/browser-tests/src/fixtures/test-db-custom.sql |
  sed \
    -e "s|__stripeWebhookSecret__|${{ secrets.BEABEE_STRIPE_WEBHOOKSECRET }}|" \
    -e "s|__stripeProductId__|${{ secrets.BEABEE_STRIPE_MEMBERSHIPPRODUCTID }}|" |
  docker compose -f docker-compose.test.yml exec -T api_app yarn backend-cli database import
```

I changed `sed -e ... <file> | docker compose ...` to `cat <file> | sed -e ... | docker compose ...` as I think it makes it simpler to understand what is happening, because the formatting makes it easier to scan and essentially both files then follow the same process (cat > import)

.github/workflows/browser-tests.yml
```diff
@@ -10,9 +10,78 @@ on:
...
+      - name: Upload database dump
+        if: failure()
+        uses: actions/upload-artifact@v4
+        with:
+          name: database
+          path: ./post-playwright-dump.sql
+
+      - name: Export Docker logs
+        if: failure()
+        run: docker compose -f docker-compose.test.yml logs >> "./docker-compose.log"
```
wpf500: What are the advantages of putting this into a file compared to just echoing it to the step? (i.e. removing `>> "./docker-compose.log"`

apps/browser-tests/src/tests/joinFlow.spec.ts
```diff
@@ -0,0 +1,222 @@
...
+test.beforeAll(async () => {
+  contributionDate = new Date().toLocaleString("en-US", {
+    month: "long",
+    year: "numeric",
+  });
+});
+
+test("Join Flow", async ({ page, browserName }) => {
+  let confirmationLink: string | null;
+  let emailAddress = getUniqueEmailAddress(browserName);
```
wpf500: Can also be a `const`?

apps/browser-tests/run.sh
```diff
@@ -0,0 +1,89 @@
...
+echo "🚀 Starting Docker Compose environment (project: $COMPOSE_PROJECT_NAME)..."
+docker compose -f docker-compose.test.yml up -d --build
+
+# Wait for services to be ready
+await_service_ready "api_app" "Server is ready and listening on port 3000" || exit 1
+#await_service_ready "frontend" "signal process started" || exit 1
+
+echo "🧪 Importing test data..."
+docker compose exec -T api_app yarn backend-cli database import < ./apps/browser-tests/src/fixtures/test-db-default.sql
+docker compose exec -T api_app yarn backend-cli database import < ./apps/browser-tests/src/fixtures/test-db-custom.sql
```
wpf500: I guess this command needs to be updated to include the `sed` replacements?

apps/browser-tests/run.sh
```diff
@@ -0,0 +1,89 @@
...
+exec > >(tee -a "$LOG_DIR/test.log") 2>&1
+
+# Load environment variables
+set -a
+source .env
+source .env.test
+set +a
+
+# Set a unique project name for e2e tests to avoid conflicts
+COMPOSE_PROJECT_NAME="beabee-e2e-tests"
```
wpf500: Should be `beabee-browser-tests` or something so it doesn't conflict with the e2e tests

.github/workflows/browser-tests.yml
```diff
@@ -10,9 +10,78 @@ on:
...
 jobs:
   run-browser-tests:
-    if: github.event_name == 'workflow_dispatch' || github.event_name == 'push' || github.event.review.state == 'approved'
+    if: >
+      github.event_name == 'push' ||
+      github.event_name == 'workflow_dispatch' ||
+      (github.event_name == 'pull_request_review' && github.event.review.state == 'approved')
     runs-on: ubuntu-latest
+    env:
+      COMPOSE_PROJECT_NAME: beabee-e2e-tests
```
wpf500: Should be `beabee-browser-tests` for consistency I think

wpf500 (changes requested):
Great to see that we have our first browser test :tada:. A few changes needed I think but overall👌

apps/e2e-api-tests/vitest.config.ts
aasthas9: Ah, I didn't notice this tbh. The file deletion came from your test setup commit that I cherry-picked (3a8a337d) 

Do we use vitest? I would assume this was deleted because we're not using vitest or vitest/docker-compose-setup anymore.

apps/frontend/docker-entrypoint.sh
aasthas9: Now that I know everything else works, I'll try the multiline format again.

.github/workflows/browser-tests.yml
aasthas9: Putting it in a log file enables exporting these logs as artifacts

apps/browser-tests/run.sh
aasthas9: This script is not used anymore. I kept it in case it is needed for the API tests later.

docker-compose.test.yml
aasthas9: Shouldn't be. It's from my original attempt to access the test dump from a bind mount. Removed it.

.github/workflows/browser-tests.yml
aasthas9: I'll keep it as `e2e` since all the tests will get integrated into this workflow?

apps/browser-tests/run.sh
aasthas9: I've removed this script for now since it isn't used for anything

.github/workflows/browser-tests.yml
aasthas9: Good idea! Thanks. :)

.github/workflows/browser-tests.yml
wpf500: I think it would also be useful for the logs to be in GitHub action UI, so you can also just look at them there as you can with other logs.

This could be solved with `tee`, which prints input to stdout and a file at the same time

```bash
docker compose -f docker-compose.test.yml logs | tee docker-compose.log
```

wpf500 (changes requested):

wpf500 (dismissed):

wpf500:
Ah @aasthas9 the Test on Approval action fails here because it's trying to run `yarn test` on all the apps/packages, which is actually trying to run the browser tests!

wpf500:
I think it might be worth excluding the browser tests from the global `yarn test` command (in `ROOT/package.json`). I think the `--exclude` flag will do the job: https://yarnpkg.com/cli/workspaces/foreach

aasthas9:
@wpf500 yep, `--exclude` does the trick :) thanks!

wpf500 (approved):

PR #598, opened 2026-05-12
feat: Add environment variable to disable rate limiting

aasthas9:
## Add env variable to disable rate limiting

`BEABEE_DISABLE_RATE_LIMIT` disables rate limiting for *all* API endpoints.

### DIFF
--- a/packages/core/src/config/config.ts
+++ b/packages/core/src/config/config.ts
@@ -263,6 +263,9 @@ export const config = {
   logging: env.ss('BEABEE_LOGGING', []), // Enable logging (default: false)
   logFormat: env.e('BEABEE_LOGFORMAT', ['json', 'simple'] as const, 'json'), // Log format (default: json)
 
+  // Rate limiting configuration
+  disableRateLimit: env.b('BEABEE_DISABLE_RATE_LIMIT', false), // Disable rate limiting in test mode (default: false)
+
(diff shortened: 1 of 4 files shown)


aasthas9:
@wpf500 do I need to change any of the checked-in example env files?

packages/core/src/config/config.ts
```diff
@@ -263,6 +263,9 @@ export const config = {
   logging: env.ss('BEABEE_LOGGING', []), // Enable logging (default: false)
   logFormat: env.e('BEABEE_LOGFORMAT', ['json', 'simple'] as const, 'json'), // Log format (default: json)
 
+  // Rate limiting configuration
+  disableRateLimit: env.b('BEABEE_DISABLE_RATE_LIMIT', false), // Disable rate limiting in test mode (default: false)
```
wpf500: By convention we don't use `_` between words (it's reserved for nested objects). Looks a bit ugly but we should stick to a common format:

TL;DR this should be `BEABEE_DISABLERATELIMIT`

wpf500 (changes requested):

wpf500:
@aasthas9 yup it would be good if you could add this to `.env.example` so it's documented somewhere. You could add it already commented out (given that it has a sensible default)

wpf500 (approved):

PR #599, opened 2026-05-13
fix(frontend): stay in current bucket after changing a response's bucket

libby-correctiv:
Previously, after the user moved a CrowdNewsroom response to a different bucket, the user stayed on that response's page, meaning they changed bucket with the response. Added routing to the previous response in the original bucket, in order that the user stays in the same bucket.

(Also a small eslint change that appeared in `yarn check`)

## Checklist before merging

### DIFF
--- a/apps/frontend/src/pages/admin/crowdnewsroom/view/[id]/responses/[rid].vue
+++ b/apps/frontend/src/pages/admin/crowdnewsroom/view/[id]/responses/[rid].vue
@@ -192,7 +192,7 @@ import {
 } from '@fortawesome/free-solid-svg-icons';
 import { computed, ref, toRef, watchEffect } from 'vue';
 import { useI18n } from 'vue-i18n';
-import { useRoute } from 'vue-router';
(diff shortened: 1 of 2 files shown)


apps/frontend/src/pages/admin/crowdnewsroom/view/[id]/responses/[rid].vue
```diff
@@ -268,7 +269,11 @@ async function handleUpdate(
   doingAction.value = true;
   try {
     await client.callout.response.update(response.value.id, data);
-    await refreshResponse();
+
+    const prevResponseId = prevResponse.value?.id;
```
Comment (wpf500) [inline apps/frontend/src/pages/admin/crowdnewsroom/view/[id]/responses/[rid].vue]: I guess this will be `undefined` if there is no previous response. I suggest to use the `nextResponse` if `prevResponse` is empty, and if both are empty then either stay with the current response (i.e. bring back `refreshResponse`) or go back to the bucket response list.

wpf500 (changes requested):

wpf500 (approved):

PR #601, opened 2026-05-18
feat(db-report): generate database reports using schemaspy

aasthas9:
## feat: Generate database reports using Schemaspy

Added new workflow that uses the test Docker stack to start only DB and migration containers. Once the containers are up, it uses [v7.0.2 of Schemaspy](https://hub.docker.com/r/schemaspy/schemaspy/tags) to generate a report, including relationship diagrams.

The workflow is set to trigger automatically only on push events to `main`. 

### DIFF
--- a/.github/workflows/database-report.yml
+++ b/.github/workflows/database-report.yml
@@ -5,14 +5,51 @@ on:
     branches:
       - main
   workflow_dispatch:
-  pull_request_review:
-    types: [submitted]
 
 jobs:
   generate-db-report:
     runs-on: ubuntu-latest
-    if: github.event_name == 'workflow_dispatch' || github.event_name == 'push' || github.event.review.state == 'approved'
 
     steps:
-      - name: Say hello
-        run: echo "Hello world"
+      - name: Checkout repository
+        uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


.github/workflows/database-report.yml
```diff
@@ -5,14 +5,51 @@ on:
...
+        run: docker compose -f docker-compose.test.yml -p beabee-db-schema up -d --build db migration
+
+      - name: Generate schemaspy report
+        run: |
+          mkdir -p db-report
+
+          docker run --rm --user root \
+            --network beabee-db-schema_default \
+            -v "$(pwd)/db-report:/output" \
+            schemaspy/schemaspy:latest \
```
wpf500: Generally I think it's good practice to pin versions instead of using `latest` (or other similar tags), this prevents the pipeline from randomly breaking in the future if latest gets some breaking updates. In this case [7.0.2](https://hub.docker.com/r/schemaspy/schemaspy/tags) seems like the actual version here.

wpf500 (changes requested):
This is great! One small change request, otherwise ready to go

wpf500 (approved):

PR #603, opened 2026-05-21
ci(api-tests): Add workflow to run end-to-end API tests

aasthas9:
### Major changes

1. Separated setup and cleanup actions for tests
`test-setup` to set up Docker environment and database, `test-teardown` to export docker logs and DB
2. New workflow for API tests
3. Added API key anonymizer
API key table is anonymized and exported for export levels `full` and `test`
4. Changed default test dump
API tests require two known users - test and rate-limit-test - as well as API keys and some payments. This is now part of the default data dump.
5. Replaced `.env.test` with fixture
Added a fixture that contains API information (path, port) and authentication information for test users

### Minor changes
1. Added unique ID to Docker project name
2. In browser tests, replaced `localhost + ROUTER_PORT` with `BEABEE_AUDIENCE`


aasthas9:
@wpf500 I assume `yarn test` will fail here as well. Should I exclude `e2e-api-tests` as we did with browser tests?

apps/e2e-api-tests/src/tests/client/callout-response-segment.test.ts
```diff
@@ -1,5 +1,8 @@
 import { BeabeeClient } from '@beabee/client';
-import { API_KEY, HOST, PATH } from '@beabee/test-utils/vitest/env';
+import {
+  api,
+  testUser,
+} from '@beabee/e2e-api-tests/src/fixtures/api-test-info.json';
```
wpf500: These imports should either be relative to the file or internal to the package. In general there are two warning signs here that indicate something could be improved:
1. Using absolute imports to reference something within a package itself
2. Importing from the `src/` directory

```ts
import { api, testUser } from '../../fixtures/api-test-info.json'; // OKAY - relative to file
import { api, testUser } from '#fixtures/api-test-info.json'; // BETTER - internal package import
```

The latter is better because `../..` is very brittle (i.e. if the file gets moved it breaks), creates a more "hidden" dependency between the files, and is "too powerful" (it could be used to go outside of the package, e.g. `../../../common/...`). There are unfortunately lots of examples where we use `../` but they should really all be removed.

Explicitly declaring and using `#fixtures/...` makes the dependency very clear (it has to be declared in the `package.json` `imports` directive).

Note: this needs to be changed in a few files.

wpf500 (changes requested):
Great to see this consolidation 🎉.

I think it should be possible to delete the whole of `packages/test-utils`. There is one dependency on the method `createTestFile` which is being used in the API tests, but I would move this into e.g. `apps/e2e-api-tests/src/utils/file.ts`.

We will almost certainly need some shared test package between the browser and API tests in the future, but I think it's better to add it back when we need it.

apps/e2e-api-tests/src/utils/file.ts
wpf500, line 1: I don't think these two new imports are needed are they?

wpf500 (changes requested):

apps/e2e-api-tests/src/utils/file.ts
aasthas9: These imports were in the vitest setup script earlier. The other imports (`fs` and `path`) do not work if these two are taken out.

wpf500: This means the application is relying on a side-effect of the `dotenv` package as `config` is unused,  so it's better to import the package without exposing methods/properties as it makes it clear it's for side effects: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import#import_a_module_for_its_side_effects_only

```ts
import 'dotenv';
```

I find it however really strange that the method in this file needs this import. I suspect it's actually being used somewhere else and this is just "accidentally working", and it would be better for the import to be moved to the place where it is actually doing something.

Let me do a bit of investigation here and get back to you... 🔍

wpf500: Resolved with #615

PR #607, opened 2026-05-27
feat(frontend): add option to select all items in table

libby-correctiv:
Adding the option to select all items in a selectable paginated table, rather than just those displayed on the current page. Implemented via a banner that appears when all items on the current page are selected, giving the option to select all items in the table.

Also adding persistent selection state across pagination so that selections are preserved when navigating between pages.

Implemented in both existing selectable tables - the contacts table and the callout responses table

## Describe your changes

- Changed selection from being stored on individual items (`item.selected`) to a separate state object `selectionState`
  - `explicit` mode: stores selected IDs
  - `all` mode: treats all items as selected except excluded IDs
- in `index.vue`, removed reapplication of selection on items after refresh, no longer needed due to new state object
- Added `useSelectionState` composable to manage selection logic
- Added `SelectAllBanner` component to control `selectionState.mode`
- Updated `AppTable` to use a controlled `selectedIds` array instead of item-level state
- Updated `AppPaginatedTable` to own `selectionState` and pass:
  - `selectedIds` to `AppTable`
  - `mode` to `SelectAllBanner`
  - Update `selectionState` based on user interactions in child components
- Added indeterminate state to `AppCheckbox`

### Screenshots

Screenshot 2026-06-01 at 12 32 17

Screenshot 2026-06-01 at 12 32 27

## Checklist before merging

not yet - will add once the english copy is approved


packages/vue/src/components/form/AppCheckbox.vue
```diff
@@ -81,7 +89,13 @@ const emit = defineEmits<{
   (e: 'update:modelValue', value: boolean): void;
 }>();
 
+const value = computed({
```
wpf500: Since this component was created Vue has introduced `defineModel` for this purpose, so this could be replaced with something like

```ts
const value = defineModel<boolean>();
```

packages/vue/src/components/form/AppCheckbox.vue
```diff
@@ -58,6 +64,8 @@ import { computed, ref, toRef, watch } from 'vue';
 export interface AppCheckboxProps {
   /** Current checked state of the checkbox */
   modelValue?: boolean;
+  /** Whether the checkbox is indeterminate */
```
wpf500: Could this be combined with `modelValue`? A checkbox can now be `true`, `false` or indeterminate, but having a separate prop implies these are independent states.

I think you could model this more accurately by changing the type of `modelValue` to something like `boolean | 'indeterminate'`. You'll need to make a few updates in the component (e.g. `value === true` instead of `value` conditions), but I think the abstraction is clearer

apps/frontend/src/composables/useSelectionState.ts
```diff
@@ -0,0 +1,67 @@
+import type { RuleGroup } from '@beabee/beabee-common';
+
+import { type ComputedRef, type Ref, computed, ref } from 'vue';
+
+export type SelectionState =
```
wpf500: Types should generally be in `src/type`. There are definitely cases where this didn't happen in the past, but those are exceptions that should be ignored :)

apps/frontend/src/composables/useSelectionState.ts
```diff
@@ -0,0 +1,67 @@
...
+  | { mode: 'explicit'; ids: string[] }
+  | { mode: 'all'; excludedIds: string[] };
+
+export function useSelectionState<I extends { id: string }>(
```
wpf500, line 17: In both uses of `useSelectionState` the code is the same:

```ts
useSelectionState(
  computed(() => contactsTable.value?.items ?? []),
  computed(() => contactsTable.value?.total ?? 0)
);
```

I can see an argument for keeping `useSelectionState` generic, but I think you could make a simple wrapper method which handles `Paginated` types as this will be the most common pattern across the code base.

Something like this?

```ts
export function usePaginatedSelectionState<Item>(
  table: Ref<Paginated<Item> | undefined>
) {
  return useSelectionState(
    computed(() => table.value?.items ?? []),
    computed(() => table.value?.total ?? 0)
  );
}
```

apps/frontend/src/composables/useSelectionState.ts
```diff
@@ -0,0 +1,67 @@
...
+  items: Ref<I[]> | ComputedRef<I[]>,
+  totalItemCount: Ref<number>
+) {
+  const selectionState = ref<SelectionState>({
+    mode: 'explicit',
+    ids: [],
+  });
+
+  function isSelected(id: string): boolean {
+    if (selectionState.value.mode === 'explicit') {
```
wpf500: This is a personal preference but I like using exhaustive switch statements in cases like this:

```ts
switch (selectionState.value.mode) {
  case 'explict':
    return selectionState.value.ids.includes(id);
  case 'all':
    return !selectionState.value.excludedIds.includes(id);
}
```

Exhaustive means the switch statement covers all possible values (in this case all and explicit). What's good about this pattern is that if in the future someone adds a new mode, the compiler will tell them they need to think about how to handle it in these methods.

```ts

// Add 'implicit' as selection state mode

function isSelected(id: string): boolean {
  switch (selectionState.value.mode) {
    case 'explict':
      return selectionState.value.ids.includes(id);
    case 'all':
      return !selectionState.value.excludedIds.includes(id);
  }
  // Error: 'implicit' not handled
}

function isSelected(id: string): boolean {
  if (selectionState.value.mode == 'explicit') {
    return selectionState.value.ids.includes(id);
  }

  // Both 'implicit' and 'all' use this. Is that intended?
  return !selectionState.value.excludedIds.includes(id);
}

File: apps/frontend/src/pages/admin/contacts/index.vue [hunk 6]
```diff
@@ -417,26 +416,14 @@ async function refreshResponses() {
...
 
-    // Preserve selection state for existing contacts
-    contactsTable.value = {
-      ...newContacts,
-      items: newContacts.items.map((c) => ({
-        ...c,
-        selected: selectedIds.has(c.id),
-      })),
-    };
+    contactsTable.value = newContacts;
```
wpf500: Is there still a need for `newContacts` or can you just do `contactsTable.value = await client.contact.list(..)`?

apps/frontend/src/pages/admin/crowdnewsroom/view/[id]/responses/index.vue
```diff
@@ -542,13 +535,7 @@ async function refreshResponses() {
...
     );
 
-    responses.value = {
-      ...newResponses,
-      items: newResponses.items.map((r) => ({
-        ...r,
-        selected: selectedIds.has(r.id),
-      })),
-    };
+    responses.value = newResponses;
```
Comment (wpf500) [inline apps/frontend/src/pages/admin/crowdnewsroom/view/[id]/responses/index.vue]: Can `newResponses` be replaced with direct assignment to `responses.value`?

wpf500 (changes requested):
This is a great PR that introduces a new feature and cleans up the previously hackier way of handling selection state 🎉.

A few changes requested but overall really impressive.

packages/vue/src/components/form/AppCheckbox.vue
libby-correctiv: Ah yeah good point that is better!

apps/frontend/src/composables/useSelectionState.ts
libby-correctiv: Looks good have added :)

apps/frontend/src/composables/useSelectionState.ts
libby-correctiv: Ah yeah that looks a lot cleaner too. I’ve changed them but in the `selectedCount` computation eslint was complaining so I had to add `default: return 0;` - annoyingly defeats what you said above in that case

packages/vue/src/components/form/AppCheckbox.vue
```diff
@@ -17,19 +17,19 @@
...
         class="flex h-5 w-5 items-center justify-center rounded border bg-white transition-colors focus:outline-none focus:ring-2 focus:ring-primary-40 focus:ring-offset-2"
         :class="[
-          value ? borderVariantClasses[variant] : 'border-grey-dark',
-          !disabled && !value ? hoverVariantClasses[variant] : '',
+          value !== false ? borderVariantClasses[variant] : 'border-grey-dark',
+          !disabled && value === false ? hoverVariantClasses[variant] : '',
           disabled ? 'cursor-not-allowed opacity-50' : '',
         ]"
       >
         <font-awesome-icon
```
wpf500, line 25: Slightly cleaner (less duplication):

```vue
<font-awesome-icon
  v-if="value"
  :icon="value === 'indeterminate' ? faMinus : faCheck"
  class="h-4 w-4"
  :class="iconVariantClasses[variant]"
/>
```

apps/frontend/src/components/table/AppPaginatedTable.vue
```diff
@@ -85,6 +94,90 @@ defineProps<{
...
+  get: () => {
+    if (selectionState.value.mode === 'all') {
+      const excludedIds = selectionState.value.excludedIds;
+      return allIdsOnPage.value.filter((id) => !excludedIds.includes(id));
+    } else {
+      const ids = selectionState.value.ids;
+      return allIdsOnPage.value.filter((id) => ids.includes(id));
+    }
+  },
+  set: (val) => {
```
wpf500: This is correct but slightly mind bending logic. I wonder how we can make this easier to read.

What do you think about?

```ts
set: (newSelectedIds) => {
  if (selectionState.value.mode === 'all') {
    const otherExcludedIds = selectionState.value.excludedIds.filter(
      (id) => !allIdsOnPage.value.includes(id)
    );
    const newExcludedIds = allIdsOnPage.value.filter(
      (id) => !newSelectedIds.includes(id)
    );
    selectionState.value = {
      mode: 'all',
      excludedIds: [...otherPageExcludedIds, ...newExcludedIds]
    }
  } else {
    const otherSelectedIds = selectionState.value.ids.filter(
      (id) => !allIdsOnPage.value.includes(id)
    );
    selectionState.value = {
      mode: 'explicit',
      ids: [...otherSelectedIds, ...newSelectedIds]
    }
  }
}
```

My thinking:
* Rename `val` to `newSelectedIds`
  `val` is fine for small computed methods, but once it gets a bit more complex it's worth giving the variable a better name
* Move all the filtering to before `selectionState.value = ...`
  This out means there are less things happening in one statement and it separates the "preparation" from the "setting"
* Remove `newIds`
  I'm fairly sure this is redundant (`newIds === newSelectedIds`)

This is mostly subjective though, so definitely interested in what you think. Probably a comment summarising what the setter is trying to do would be helpful too (e.g. preserving IDs from other pages while updating those within this page).

wpf500 (changes requested):
One change I think is defo worth it (the cleaner icon code), the other is just a suggestion.

apps/frontend/src/composables/useSelectionState.ts
wpf500: Yeah this is a known problem with the Vue linter, there are a few other places in the codebase that have to do this.

However I've just seen that someone has merged [a fix](https://github.com/vuejs/eslint-plugin-vue/pull/3067) which looks like it would fix this, so maybe we can remove all these cases soon

apps/frontend/src/components/table/AppPaginatedTable.vue
libby-correctiv: Ah this is definitely way cleaner, looks good to me :) I've made all those changes and added a comment

packages/vue/src/components/form/AppCheckbox.vue
libby-correctiv: Oh oops ..... me not checking claude's work

libby-correctiv: changed

wpf500 (dismissed):

wpf500 (approved):

PR #608, opened 2026-05-27
test(browser-tests): Add browser test for one-time contribution

aasthas9:
## test: Add browser test for one-time contribution

1. Added more information to browser test fixture: test card details, login credentials for known user
2. New test logs in (with known-user credentials), makes one-time contribution, then verifies that the payment shows up under "contributions"

### DIFF
--- a/apps/browser-tests/src/tests/one-time-contribution.ts
+++ b/apps/browser-tests/src/tests/one-time-contribution.ts
@@ -1,5 +1,5 @@
 import { test, expect } from "@playwright/test";
-import { cardInfo, signIn } from "#fixtures/user-info.json";
+import { cardInfo, signIn } from "#fixtures/testData.json";
 
 const email: string = signIn.email;
 const testPw: string = signIn.password;
--- a/apps/browser-tests/src/tests/oneTimeContribution.spec.ts
+++ b/apps/browser-tests/src/tests/oneTimeContribution.spec.ts
@@ -56,12 +56,6 @@ test("One-time contribution after login", async ({ page }) => {
     .getByText("Thank you for your contribution!")
     .waitFor({ state: "visible" });
 
-  await page.reload();
-  await expect(
-    page.getByRole("heading", { name: "Payment history" }),
-    "Payment history visible",
-  ).toBeVisible();
-
   const targetRow = page
     .locator("tr")
     .filter({ hasText: contributionDate })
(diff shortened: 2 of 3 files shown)


apps/browser-tests/src/tests/one-time-contribution.ts
```diff
@@ -0,0 +1,73 @@
...
+    .contentFrame();
+
+  await stripeFrame.locator("#payment-numberInput").fill(cardInfo.number);
+  await stripeFrame.locator("#payment-expiryInput").fill(cardInfo.validUntil);
+  await stripeFrame.locator("#payment-cvcInput").fill(cardInfo.cvv);
+  await stripeFrame.locator("#payment-countryInput").selectOption("DE");
+  await page.getByRole("button", { name: /continue/i }).click();
+
+  await page
+    .getByText("Thank you for your contribution!")
```
aasthas9, line 56: @wpf500 would it be a good idea to import `en.json` and get the string from there?

wpf500: I think we do need to think about a strategy in general for all the hardcoded text in these tests, e.g. using special test props or something.

I wouldn't see this as a blocker for this PR as we have lots of hardcoded text, but would be a good thing to scope out and bring into an upcoming sprint.

apps/browser-tests/src/tests/oneTimeContribution.spec.ts
```diff
@@ -0,0 +1,73 @@
...
+  await stripeFrame.locator("#payment-expiryInput").fill(cardInfo.validUntil);
+  await stripeFrame.locator("#payment-cvcInput").fill(cardInfo.cvv);
+  await stripeFrame.locator("#payment-countryInput").selectOption("DE");
+  await page.getByRole("button", { name: /continue/i }).click();
+
+  await page
+    .getByText("Thank you for your contribution!")
+    .waitFor({ state: "visible" });
+
+  await page.reload();
```
wpf500: I think there is a small race condition here if the payment completed webhook from Stripe is a bit slow. Could you add a bit of retry logic here?

e.g. reload page and check for expected item, if not there wait a second then reload again (up to 5 reloads then it fails)

wpf500 (changes requested):

apps/browser-tests/src/tests/oneTimeContribution.spec.ts
aasthas9: `waitFor()` waits until default timeout (30s) for the "thank you for your contribution" pop-up message, and then reloads the page. The pop-up appears after Stripe's response, so the 30s wait should account for a slow webhook, right?

wpf500: The confirmation the user sees is separate to the payment webhook. i.e. Stripe confirms to us that the payment was successfully initiated, we say thank you to the user, then Stripe confirms the payment was successful.

In practice this "initiated -> success" process is sub-second and Stripe webhooks are very quick, but there is at least a chance that this test could fail because the page reloaded too quickly. I think a quick retry loop is worth it here just to be sure we don't get an unrelated failure somewhere down the line.

aasthas9: @wpf500 Added. I've also removed the check for the "payment history" heading since it was based on the assumption that no previous payment history exists. The test only looks for the actual payment details now.

wpf500 (dismissed):

wpf500 (approved):

PR #610, opened 2026-05-30
fix(deps): dedupe prosemirror to fix rich text line breaks

JumpLink:
## Summary

Fixes the regression reported on #578 ([comment](https://github.com/beabee-communityrm/monorepo/pull/578#issuecomment-4565045097)): rich text inputs can no longer accept line breaks. Pressing <kbd>Enter</kbd> in the membership builder or the callout / CrowdNewsroom builder throws and the line break is dropped.

```
RangeError: Can not convert <> to a Fragment
(looks like multiple versions of prosemirror-model were loaded)
```

## Root cause

The TipTap 2 → 3 bump (#578) pulls **three** copies of `prosemirror-model` into the tree — all from within TipTap's own dependencies:

| Consumer | range | resolved |
|---|---|---|
| `@tiptap/pm` | `^1.24.1` | **1.25.7** |
| `prosemirror-tables` (via `@tiptap/pm`) | `^1.25.0` | **1.25.1** |
| `@tiptap/pm` → nested `prosemirror-view` | `^1.20.0` | **1.24.1** |

`prosemirror-view` and `prosemirror-transform` were duplicated the same way. Because of `nodeLinker: node-modules`, every copy is bundled. The editor builds its schema with one `prosemirror-model` copy (`@tiptap/pm/model` → 1.25.7) but the view manipulates nodes through another (`@tiptap/pm/view` → `prosemirror-view` → 1.24.1). `Fragment.from()`'s `instanceof` check fails across copies → the `RangeError` on the first paragraph split (Enter).

PR #578's description said the four core ProseMirror packages were "pinned via root `resolutions`", but that change was never committed — the root `package.json` was untouched and the lockfile still carried the duplicates.

## Fix

Add the missing root `resolutions`, pinning each duplicated core package to the single version `@tiptap/pm@3.23.5` already resolves to. `prosemirror-state` was already a single copy, so it needs no pin.

```json
"prosemirror-model": "1.25.7",
"prosemirror-transform": "1.10.4",
"prosemirror-view": "1.41.8"
```

## Verification

Reproduced and verified locally against the `@beabee/vue` `AppRichTextEditor` story (histoire + Playwright):

| | before (`main`) | after |
|---|---|---|
| `prosemirror-model` copies on disk | 6 | **1** |
| copies bundled into `@beabee/vue` | 3 | **1** |
| Press <kbd>Enter</kbd> | ❌ `RangeError … multiple versions of prosemirror-model` | ✅ no error |
| Line break | dropped — `<p>…AB</p>` | inserted — `<p>…A</p><p>B</p>` |

image

- [x] `yarn why prosemirror-model` → single `1.25.7`
- [x] `@beabee/vue` `check` (vue-tsc + prettier) ✅
- [x] `@beabee/vue` `build:lib` ✅ (bundle −113 kB from dropping the duplicate copies)
- [x] Live editor test: Enter inserts a line break, console clean

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,6 +35,9 @@
     "@types/express-serve-static-core": "^5.1.0",
     "@types/express-session": "^1.19.0",
     "brace-expansion@^5.0.5": "^5.0.6",
+    "prosemirror-model": "1.25.7",
+    "prosemirror-transform": "1.10.4",
+    "prosemirror-view": "1.41.8",
     "tar-fs": "^3.1.1"
   },
   "dependenciesMeta": {
--- a/yarn.lock
+++ b/yarn.lock
@@ -17593,30 +17593,12 @@ __metadata:
   languageName: node
   linkType: hard
 
-"prosemirror-model@npm:^1.0.0, prosemirror-model@npm:^1.20.0, prosemirror-model@npm:^1.21.0":
-  version: 1.24.1
-  resolution: "prosemirror-model@npm:1.24.1"
-  dependencies:
-    orderedmap: "npm:^2.0.0"
-  checksum: 10c0/d09e1aac3cbd451e6a8f3927df487869b1ce0b7f21836c0c6966d401f4ab2d5984988f74d3ff98c72484a78498b9c0c3ad62cf2a06ac869664f1820c97313054
-  languageName: node
-  linkType: hard
-
-"prosemirror-model@npm:^1.24.1":
+"prosemirror-model@npm:1.25.7":
   version: 1.25.7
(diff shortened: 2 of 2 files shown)


wpf500:
Using `resolutions` should always be a last resort, it silently overrides package versions and can create bigger problems in the future.

[Claude suggested](https://claude.ai/chat/8d020a19-9460-4b7f-a173-2386fa6c9299) `yarn dedupe` as a better solution in this case. I didn't know about this command before but this is exactly what we want in this case. I've applied this in #612 so we can do a release and am closing this PR.

PR #615, opened 2026-06-02
build(tsconfig): fix dotenv dependency in tests

wpf500:
This PR fixes the strange problem where importing `dotenv` made the tests compile, even though it wasn't used.

I have to be honest I don't fully understand the fix. It has something to do with a side effect from the import which was putting TSC in "Node compilation mode". To fix it properly the `e2e-api-test` package needed missing Node types (`@types/node`) and for them to be loaded in the TSConfig file.

I also removed the `dotenv` dependency from a few places that were actually never using it


aasthas9 (approved):

aasthas9:
@wpf500 you can ignore the failing browser tests for now. it's just the missing playwright version bump

PR #617, opened 2026-06-04
feat: Newsletter integrations tab

libby-correctiv:
## feat (integrations tab): Add new tab for newsletter integrations

Adds a new Integrations tab to admin settings page that displays the health status and other details of newsletter providers. Each provider is shown as a card - disabled if not configured, otherwise reflecting live status from a new API endpoint.

Currently only configured for Mailchimp.

#### Frontend (Libby)

- Added new integrations tab in settings page
- Added integration card components: `IntegrationCard`, `IntegrationCardHeader` (displays provider icon, name, audience ID, status pill, refresh button), `MailchimpCardContent` (groups table), `StatusPill`
- New Integration discriminated union type (`DisabledIntegration` | `MailchimpIntegration`) mirroring the common types plus any props needed for display (eg icon, color)
- Added `useNewsletterIntegrations` composable which constructs the data to be displayed in the cards, merging a `providerMap` with static info with the API data. Also handles refresh logic in a function that calls the refresh endpoint and shows a  notification based on the health transition (connected/disconnected/reconnected/no changes).
- Changed contact list newsletter groups filter to use the integrations API instead of join/setup page content

#### Backend (Aastha)
- Use `Equals` instead of `IsIn` when checking newsletter provider. 
- Change `{ info }` to `info` in `plainToInstance`

### Screenshots
Screenshot 2026-06-18 at 11 56 00

Screenshot 2026-06-08 at 18 35 37

## Checklist before merging


aasthas9:
@libby-correctiv I refactored the shared object a little bit. See 2e634c6b319d44ea468bba931af33ef03a3dd41e

aasthas9:
Renamed the PR to pass linter check

apps/frontend/src/type/integration.ts
```diff
@@ -0,0 +1,29 @@
...
+}
+
+/** Any provider that is not yet configured */
+export interface DisabledIntegration extends IntegrationDisplayProps {
+  provider: string;
+  status: ApiHealthStatus.DISABLED;
+}
+
+/** Mailchimp provider, active with full audience/group data */
+export interface MailchimpIntegration extends IntegrationDisplayProps {
```
wpf500: I think you could reduce duplication here and use the `@beabee/common` type:

```ts
export interface MailchimpIntegration extends IntegrationDisplayProps, MailchimpNewsletterIntegrationData {}
```

apps/backend/src/api/dto/NewsletterIntegrationDto.ts
```diff
@@ -0,0 +1,37 @@
...
+  NoneNewsletterIntegrationData,
+} from '@beabee/beabee-common';
+
+import { Type } from 'class-transformer';
+import { IsEnum, IsIn, IsString, ValidateNested } from 'class-validator';
+
+import { NewsletterGroupDto } from './NewsletterDto.js';
+
+export class NoneNewsletterIntegrationDto implements NoneNewsletterIntegrationData {
+  @IsIn(['none'])
```
wpf500: Can be more simply expressed with `@Equals('none')`. Applies to a few other places in this file

wpf500 (changes requested):
I think the general direction here is good, but some of the over generalisation of integrations in the frontend makes this more complicated than it probably needs to be.

In general it's always good to consider if it's worth generalising something when there is only one use case. Clearly here we will want to build out this page so there is a good underlying motivation, but I'd like to chat through some of the frontend components to see if we can simplify things down.

apps/frontend/src/composables/useNewsletterIntegrations.ts
```diff
@@ -0,0 +1,84 @@
...
+  const integrations = ref<Integration[]>([]);
+  const loading = ref(false);
+  const error = ref<Error | null>(null);
+
+  async function load() {
+    loading.value = true;
+    error.value = null;
+    try {
+      // Build all known provider cards upfront as disabled
+      const result: Integration[] = Object.keys(providerMap).map(
```
wpf500:
```ts
const data = await client.integrations.getNewsletter();

integrations.value = Object.keys(providerMap).map(provider => {
  return data.provider === provider ? buildIntegration(data) : buildDisabledIntegration()
});
```

apps/frontend/src/pages/admin/settings/integrations.vue
```diff
@@ -0,0 +1,66 @@
...
+          t(`adminSettings.integrations.categories.${category}`)
+        }}</AppSectionHeading>
+
+        <div class="space-y-4">
+          <IntegrationCard
+            v-for="integration in items"
+            :key="integration.provider"
+            :integration="integration"
+          >
+            <MailchimpCardContent
```
wpf500: Assume all newsletter integrations are the same (so more like `NewsletterCardContent`)

apps/frontend/src/components/integrations/IntegrationCardHeader.vue
```diff
@@ -0,0 +1,54 @@
...
+        v-if="integration.icon"
+        :icon="integration.icon"
+        class="text-3xl"
+      />
+      <span v-else>{{ integration.name[0] }}</span>
+    </div>
+    <div class="min-w-0 flex-1">
+      <p class="font-semibold">{{ integration.name }}</p>
+      <p v-if="'audienceId' in integration" class="text-sm text-body-80">
+        {{ t('adminSettings.integrations.mailchimp.audienceId') }}:
```
wpf500: Different namespace, not `mailchimp`

apps/frontend/src/pages/admin/settings/integrations.vue
```diff
@@ -0,0 +1,66 @@
...
+import MailchimpCardContent from '#components/integrations/MailchimpCardContent.vue';
+import { useNewsletterIntegrations } from '#composables/useNewsletterIntegrations';
+import { ApiHealthStatus } from '@beabee/beabee-common';
+
+const { t } = useI18n();
+
+const { integrations, loading, load } = useNewsletterIntegrations();
+
+const integrationsByCategory = computed(() =>
+  integrations.value.reduce<Record<string, typeof integrations.value>>(
```
wpf500: Reuse `groupBy` if it exists or create this in a util method

wpf500 (changes requested):

apps/frontend/src/composables/useNewsletterIntegrations.ts
```diff
@@ -0,0 +1,76 @@
...
+  const error = ref<Error | null>(null);
+
+  async function load() {
+    loading.value = true;
+    error.value = null;
+    try {
+      const data = await client.integrations.getNewsletter();
+
+      integrations.value = Object.keys(providerMap).map((provider) => {
+        return data.provider !== 'none' && data.provider === provider
```
wpf500, line 68: I think you don't even need to check if `data.provider !== 'none'` here as `provider` can never be none and this simplifies the condition (easier to understand)

apps/frontend/src/pages/admin/settings/integrations.vue
```diff
@@ -0,0 +1,62 @@
...
+    </template>
+  </div>
+</template>
+
+<script lang="ts" setup>
+import { AppLoadingSpinner, AppSectionHeading } from '@beabee/vue';
+import { computed, onMounted } from 'vue';
+import { useI18n } from 'vue-i18n';
+
+import { groupBy } from '@beabee/core/utils/objects';
```
wpf500: `@beabee/core` can only be used by backend services, not the frontend. This hopefully shouldn't even compile as the frontend and backends use very different environments.

wpf500 (changes requested):

apps/frontend/src/composables/useNewsletterIntegrations.ts
libby-correctiv: Ah yeah it is redundant logically but I had to add it to let typescript know that data will be type MailchimpNewsletterIntegrationData instead of NoneNewsletterIntegrationData (because buildNewsletterIntegration expects MailchimpNewsletterIntegrationData)

wpf500: Ah I see, makes sense 👍

wpf500 (approved):

PR #620, opened 2026-06-08
chore: trying to release v0.45.0

wpf500:
I tried to use the new `release-it` workflow but hit a number of problems. I eventually managed to work my way through most of them but hit one blocker at the end so I released v0.45.0 using our existing release process. I then used `release-it` just to bump the version numbers, which is included in this PR

### Changelog still not generating correctly

Merge commits were not being included, non-merge commits were. I fixed this by updating `gitRawCommitsOpts`, but some merges were still missing. I think this is because the new merge commit settings were only introduced halfway through this release. I hope this will go away now all future PRs will be merged with the expected commit format.

### Sections for changelog seemingly random

I removed the `section` configuration as we don't currently use this, and we should discuss what it means before adding it. Changelogs were then generated in line with current standards.

### Unable to push to `main`

I don't know what the best way to resolve this is. We have `main` branch protection on so `release-it` can't actually create a release commit.

### DIFF
--- a/.release-it.json
+++ b/.release-it.json
@@ -13,7 +13,6 @@
     "publish": false
   },
   "hooks": {
-    "before:init": ["yarn check", "yarn build"]
   },
   "plugins": {
     "@release-it/bumper": {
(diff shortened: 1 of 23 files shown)


wpf500:
I asked Claude if it was possible to override the branch protection. It said yes, but...

> That said, bypassing branch protection for release commits is a bit of an anti-pattern. The more conventional approach for your setup would be:
>
>Do the release prep (version bumps, changelog) on a branch like release/0.45.0
>Open a PR → merge it
>Run release-it on the merged commit with --no-git.push (tag only)
>
>That way the release commit goes through your normal PR flow and you never need to bypass protection rules.

So it seems to me that with branch protection `release-it` should really be pushing to a release branch, which we can then merge and tag, maybe again with `release-it`?

JumpLink:
Tried release-please as another option in #621. It handles PR-based releases natively, so no branch protection workaround needed. Single Release PR stays open with the next bumps and changelog, you merge it to ship. Happy to bin it if you'd rather stick with release-it.

Also, probably worth doing a quick v0.45.1 after this merges to get the npm publish through. The 0.45.0 one failed since the tag was pointing at unbumped main.

JumpLink (approved):

PR #625, opened 2026-06-12
feat(api): Add new API endpoint for newsletter integrations

aasthas9:
### New API endpoint
Added new endpoint at `/api/1.0/integrations/newsletter`.
1. `GET /api/1.0/integrations/newsletter` returns the name of the newsletter provider, list ID and a list of newsletter groups cached in the database
2. `GET /api/1.0/integrations/newsletter?with=health` performs an additional connectivity check. It makes a simple GET request to the newsletter provider API. If the response code is not 200, health status is set to unhealthy. Otherwise, healthy. If no provider is configured, health status is set to 'disabled'
3. `POST /api/1.0/integrations/newsletter/refresh` gets a list of configured newsletter groups from the provider. Then, it compares this list with the cached groups and returns a diff + status information. If groups have been added/deleted, it also updates the DB cache.

### Rename and remove DB columns
Renamed `mailchimp-newsletter-groups` as `newsletter-groups`. Also moved the user-configured NL groups from `options/newsletter-groups` to `content/join/setup.data.newsletterGroups`

### Cache newsletter groups in database
Added command to setup mailchimp integrations and cache NL groups in the `newsletter-groups` column.
Usage: `yarn start setup integrations mailchimp`
The command is also called automatically when setupAll is requested.

### Minor browser test fixes
Removed the `mailchimp-newsletter-groups` entry in seed DB. I've also taken this chance to rename an existing test file. One-time contribution test had a missing `.spec` in its filename, causing it to not be detected.


wpf500 (changes requested):
Generally great PR and you even included changes to the CLI 🎉 

As agreed in our chat just now, changes are:
- Remove `mailchimp-newsletter-groups` entirely
- Make `newsletter-groups` the now-updated source of truth for groups
- Move the join form list into the `content` table and create a migration to populate it

packages/core/src/lib/mailchimp.ts
```diff
@@ -275,7 +275,6 @@ export function nlContactToMCMember(
   }
 
   const groupIds = [
```
wpf500: Very small thing but this be simplified as it no longer needs the array spreading.

```ts
const groupIds = OptionsService.getJSON()...
```

wpf500 (changes requested):

packages/core/src/migrations/1781533044882-RemoveMailchimpNLGroups.ts
```diff
@@ -0,0 +1,39 @@
...
+    await queryRunner.query(`
+      INSERT INTO content (id, updated, data)
+      SELECT 'join/setup', NOW(), jsonb_build_object('newsletterGroups', value::jsonb)
+      FROM option WHERE key = 'newsletter-groups'
+      ON CONFLICT (id) DO UPDATE SET data = content.data || EXCLUDED.data
+    `);
+
+    // Empty options/newsletter-groups so it can be used as cache store
+    await queryRunner.query(
+      `UPDATE option SET value = '[]' WHERE key = 'newsletter-groups'`
```
wpf500, line 17: Thinking about it, I think it would be simpler not to clear the cache here. The value in this option is already good enough to be used as the cache (i.e. the `id` is correctly set), and that way we can release the API changes without needing to run the group synchronisation script.#

What do you think?

wpf500 (changes requested):

packages/core/src/migrations/1781533044882-RemoveMailchimpNLGroups.ts
aasthas9: The values here would have the shape `{id: string, label: string, checked: boolean}[]`, which we do not want. And the label will be user-specified, not what's configured with the provider.

wpf500:
As per chat in standup, if you add the newsletter group refresh to the API app initialisation then this could be released today. We don't have a good place for service initialisation at the moment, `initApp` does some stuff but is used by the webhook, legacy and API apps so wouldn't be appropriate here (we just need one to update the groups, not all three!).

I would suggest just keeping it simple for now and adding it in `apps/backend/src/api/app.ts`, something like...
```ts
initApp()
  .then(() => newsletterService.refreshGroups())
  .then(() => {
    // ...
  });
```

wpf500 (dismissed):

wpf500 (approved):

PR #627, opened 2026-06-16
feat(frontend): Replace group ID textbox with label dropdown

aasthas9:
## Describe your changes
When adding newsletter groups to the membership builder or crowd newsroom, the user would have had to type in the group ID as configured with the newsletter provider. This PR:
1. Replaces the textbox with a dropdown
2. Replaces IDs with labels: the labels are those configred with the NL provider and cannot be overwritten by the user
3. Pre-fills label textbox with dropdown selection only when a new group is added
4. For pre-existing user-configured groups, matches the ID of the user group with the provider group, then selects the corresponding label from the dropdown.
5. Fixes minor issue with class transformation in IntegrationsController. `plainToInstance` was wrongly returning an object as a key-value pair.

Note that the client API implementation is the same as #617 

Reference tickets: [1](https://correctivdigital.openproject.com/projects/beabee/work_packages/2802) and [2](https://correctivdigital.openproject.com/projects/beabee/work_packages/2803)

### Screenshots

image
image
image

### DIFF
--- a/apps/frontend/src/components/newsletter/AppNewsletterOptInSettings.vue
+++ b/apps/frontend/src/components/newsletter/AppNewsletterOptInSettings.vue
@@ -69,7 +69,12 @@
         <template #default="{ item }">
           <!-- Group ID -->
           <div class="flex-1">
-            <AppInput v-model="item.id" :label="t('common.id')" required />
+            <AppSelect
+              v-model="item.id"
+              :label="t('common.id')"
+              :items="availableGroups"
+              required
+            />
           </div>
 
           <!-- Group Label -->
@@ -105,11 +110,19 @@ import {
   AppRepeatable,
   AppRichTextEditor,
   AppSectionHeading,
+  AppSelect,
 } from '@beabee/vue';
 
 import { useI18n } from 'vue-i18n';
 
 const { t } = useI18n();
+
+/**
+ * Groups configured with the newsletter provider
+ */
+defineProps<{
+  availableGroups: Omit<NewsletterGroupData, 'checked'>[];
+}>();
 
 /**
  * Model values for the component configuration
--- a/apps/frontend/src/components/newsletter/AppNewsletterOptInSettings.vue
+++ b/apps/frontend/src/components/newsletter/AppNewsletterOptInSettings.vue
@@ -72,7 +72,7 @@
             <AppSelect
               :model-value="item.id"
               :label="t('common.id')"
-              :items="availableGroups"
+              :items="cachedGroups"
               required
               @update:model-value="onSelectionChange(item, $event)"
             />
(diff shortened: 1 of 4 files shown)


apps/frontend/src/components/pages/admin/callouts/tabs/SettingsTab.vue
```diff
@@ -355,4 +358,10 @@ const canStartNow = computed(
     props.status === ItemStatus.Draft ||
     props.status === ItemStatus.Scheduled
 );
+
+const cachedGroups = ref<NewsletterGroupData[]>([]);
+
+onBeforeMount(async () => {
+  cachedGroups.value = await client.integrations.getNewsletterGroups();
```
wpf500: I think I would do this inside the `AppNewsletterOptInSettings` component, that way it doesn't need the prop to be given on each instance

apps/frontend/src/components/newsletter/AppNewsletterOptInSettings.vue
```diff
@@ -105,12 +111,31 @@ import {
...
 const { t } = useI18n();
 
+/**
+ * Groups configured with the newsletter provider
+ */
+const props = defineProps<{
+  cachedGroups: Omit<NewsletterGroupData, 'checked'>[];
+}>();
+
+function onSelectionChange(item: NewsletterGroupData, newId: string) {
```
wpf500, line 126: This isn't totally clear to me what it's doing. Is it setting the label if it was previously empty? I think it could use a comment and possibly you don't need `isNew` if you set `item.id` at the end? I think I would rename `group` to `newGroup` just for added clarity (`newId`/`newGroup` read like they are related)

wpf500 (changes requested):
A couple of small change requests for simplicity/clarity, but looks good generally :)

apps/frontend/src/components/pages/admin/callouts/tabs/SettingsTab.vue
aasthas9: Good point. Thanks :)

apps/frontend/src/components/newsletter/AppNewsletterOptInSettings.vue
aasthas9: It sets the label only for newly added groups, i.e. not overwriting pre-configured user labels. `group` is the cached group that matches the selected group ID.
1. Select a value in the dropdown (the selected value is the group ID, not the full object)
2. Check if this is a new group
3. If it is, find the label linked to the selected value from (1)

aasthas9: You're right about the `isNew`. I created a separate variable to make it more readable, but I've replaced it with better comments now.

wpf500 (dismissed):

wpf500 (approved):

PR #628, opened 2026-06-17
feat(backend-cli): add health checks for payment, email, and newsletter integrations

wpf500:
Builds on the existing `health` CLI command (document/image/newsletter) by adding payment and email integration checks, and deepening the newsletter check.

Ticket: https://correctivdigital.openproject.com/wp/2793

### CLI

Two new subcommands, following the existing pattern:

```
yarn backend-cli health payment   # Stripe + GoCardless
yarn backend-cli health email     # configured email provider
```

(plus `health all`, which now includes them.) Each integration reports `healthy` / `unhealthy` / `disabled` and the command exits non-zero if anything is unhealthy.

### What each check does

- **Payment** (`PaymentService.getHealthStatus`) — combines the payment providers into one status (rare for an instance to use both):
  - **Stripe**: verifies the configured membership product still exists and the webhook is registered with the expected API version (mirrors `setup integrations stripe`).
  - **GoCardless**: lightweight connectivity check (lists a customer) — kept minimal as GC is deprecated.
- **Email** (`EmailService.getHealthStatus` → provider) — `getHealthStatus()` is now an abstract method on the email `BaseProvider`, implemented per provider:
  - **SMTP**: `transporter.verify()` (connection + auth).
  - **Mandrill**: pings `/users/ping` to validate the API key.
  - **SendGrid** (our main provider): confirms the support email's domain is authenticated and valid via `/v3/whitelabel/domains`.
- **Newsletter** (Mailchimp) — extends the existing audience-reachable check to also verify the webhook is installed with the correct secret, sources (`user` + `admin` on, `api` off), and all the events the webhook handler processes. The known-events list now lives in core (`KNOWN_WEBHOOK_EVENTS`) and is shared with `apps/webhooks`.

### Misc

- Extracted a shared `getEmailDomain()` helper in `@beabee/core/utils/email`, reused by the SendGrid check and the `IsFromEmail` validator.
- All statuses use the shared `ApiHealthStatus` enum.

🤖 Generated with AI

### DIFF
--- a/packages/core/src/providers/email/SendGridProvider.ts
+++ b/packages/core/src/providers/email/SendGridProvider.ts
@@ -1,3 +1,5 @@
+import { ApiHealthStatus } from '@beabee/beabee-common';
+
 import sgMail from '@sendgrid/mail';
 
 import { SendGridEmailConfig } from '#config/config';
@@ -59,5 +61,10 @@ export class SendGridProvider extends BaseProvider {
         { resp }
       );
     }
+  }
+
+  async getHealthStatus(): Promise<ApiHealthStatus> {
+    // TODO: verify the SendGrid API key
+    return ApiHealthStatus.DISABLED;
   }
 }
--- a/packages/core/src/providers/email/SendGridProvider.ts
+++ b/packages/core/src/providers/email/SendGridProvider.ts
@@ -1,23 +1,38 @@
 import { ApiHealthStatus } from '@beabee/beabee-common';
 
 import sgMail from '@sendgrid/mail';
+import axios, { type AxiosInstance } from 'axios';
 
 import { SendGridEmailConfig } from '#config/config';
 import { log as mainLogger } from '#logging';
+import OptionsService from '#services/OptionsService';
 import type { EmailOptions, EmailRecipient, PreparedEmail } from '#type/index';
+import { getEmailDomain } from '#utils/email';
 
 import { BaseProvider } from './BaseProvider.js';
 
 const log = mainLogger.child({ app: 'sendgrid-email-provider' });
 
+interface SendGridAuthenticatedDomain {
+  domain: string;
+  valid: boolean;
+}
+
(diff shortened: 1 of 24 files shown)


packages/core/src/providers/email/SendGridProvider.ts
```diff
@@ -60,4 +77,32 @@ export class SendGridProvider extends BaseProvider {
       );
     }
   }
+
+  async getHealthStatus(): Promise<ApiHealthStatus> {
+    const domain = getEmailDomain(OptionsService.getText('support-email'));
```
aasthas9, line 82: Why is a domain check needed only for this provider?

packages/core/src/services/PaymentService.ts
```diff
@@ -255,6 +256,28 @@ class PaymentService {
     await this.provider(contact, (p) => p.updateContact(updates));
   }
 
+  /**
+   * Check the health of the payment integration. As it's rare for a single
```
aasthas9, line 260: Is it theoretically possible to use both providers?

aasthas9 (commented):
Looks good! I saw two things that weren't super clear to me and have left comments for those.

packages/core/src/providers/email/SendGridProvider.ts
wpf500: beabee uses domain-level email sender authorisation, so each instance can send emails from any address on the one domain it's been given permission to send from. This is the domain of the support email address
e.g. support-email = 'support@foo.tld', the beabee instance is configured with SendGrid to be able to send from all @foo.tld email addresses

packages/core/src/services/PaymentService.ts
wpf500: Yes but it is never happens in practice (GC is a deprecated payment provider). I thought it was simpler to stick to the "one service, one health" pattern we established than make it more complicated for an edge case.

aasthas9 (approved):

PR #633, opened 2026-06-24
feat(cron): Add new cron job to sync newsletter groups

aasthas9:
Added new action to newsletter syncing actions, to get groups from newsletter service and update cache. Also added a new cron job to call this action hourly.

Relevant ticket: https://correctivdigital.openproject.com/projects/beabee/work_packages/2907/activity

### DIFF
--- a/apps/backend-cli/src/actions/sync/newsletters/refresh-cached-groups.ts
+++ b/apps/backend-cli/src/actions/sync/newsletters/refresh-cached-groups.ts
@@ -1,0 +1,23 @@
+import { log as mainLogger } from '@beabee/core/logging';
+import { runApp } from '@beabee/core/server';
+import { newsletterService } from '@beabee/core/services';
+
+const log = mainLogger.child({ app: 'sync-newsletter-refresh-cached-groups' });
+
+export const refreshCachedGroups = async (): Promise<void> => {
+  await runApp(async () => {
+    log.info('Refreshing cached newsletter groups');
+
+    const { groupChanges } = await newsletterService.refreshNewsletterGroups();
+
+    if (groupChanges.length === 0) {
+      log.info('✅ Newsletter group cache is up to date, no changes required');
+      return;
+    }
+
+    log.info(`ℹ️ Updated cached groups with ${groupChanges.length} changes:`);
+    for (const change of groupChanges) {
+      log.info(`- ${change.action}: ${change.label} (${change.id})`);
+    }
+  });
+};
--- a/apps/backend-cli/src/actions/sync/newsletters/refresh-cached-groups.ts
+++ b/apps/backend-cli/src/actions/sync/newsletters/refresh-cached-groups.ts
@@ -2,13 +2,22 @@ import { log as mainLogger } from '@beabee/core/logging';
 import { runApp } from '@beabee/core/server';
 import { newsletterService } from '@beabee/core/services';
 
+import { SyncClearPendingStatusArgs } from '../../../types/sync.js';
+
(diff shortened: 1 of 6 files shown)


wpf500 (changes requested):
As per our chat, add a `--dry-run` option to this command for consistency

apps/backend-cli/src/actions/sync/newsletters/refresh-cached-groups.ts
```diff
@@ -2,13 +2,22 @@ import { log as mainLogger } from '@beabee/core/logging';
...
 import { newsletterService } from '@beabee/core/services';
 
+import { SyncClearPendingStatusArgs } from '../../../types/sync.js';
+
 const log = mainLogger.child({ app: 'sync-newsletter-refresh-cached-groups' });
 
-export const refreshCachedGroups = async (): Promise<void> => {
+export const refreshCachedGroups = async ({
+  dryRun,
+}: SyncClearPendingStatusArgs): Promise<void> => {
```
aasthas9: @wpf500 I recycled `SyncClearPendingStatusArgs` instead of defining another interface for this command. `SyncClearPendingStatusArgs` also has only one member - `dryRun`. Of course this comes with the caveat that another interface would be needed if this one changes.

packages/core/src/services/NewsletterService.ts
```diff
@@ -187,7 +187,11 @@ class NewsletterService {
    * change contains group id, label, and an action (added: present
    * on the provider but not cached OR removed": cached but no longer on the provider)
    */
-  async refreshNewsletterGroups(): Promise<NewsletterDiffData> {
+  async refreshNewsletterGroups(dryRun = false): Promise<NewsletterDiffData> {
+    if (dryRun) {
```
wpf500: I think this is just a duplicate log as in the CLI logs on dry run?

If this method was called with `dryRun = true` in the future it might be useful to add a log for this, but the log message would need more context as right now in the log aggregator this would just appear as "newsletter-service: DRY RUN - No changes will be made", i.e. should at least mention "refreshing newsletter groups" or something.

I would remove for now as you will revisit clear logging for this as part of the error logging ticket.

wpf500 (changes requested):

packages/core/src/services/NewsletterService.ts
aasthas9: I added it in case we called it with `dryRun` enabled from places other than the CLI. Like to test a failing upsert?

apps/backend-cli/src/actions/sync/newsletters/refresh-cached-groups.ts
wpf500: Makes sense to reuse but would be good to rename it. Maybe an LLM has a better suggestion but I'd go for something like `SyncDryRunnableArgs` or something

packages/core/src/services/NewsletterService.ts
wpf500: In that case it needs a clearer log message as in most contexts these logs will go to the log aggregator, so each log message should be self-explanatory. It would make sense in general to log that the groups are being refreshed so you could change this to something like:

```ts
log.info('Refreshing newsletter groups' + (dryRun ? ' - dry run' : ''));
```
i.e. always logging something (without the `if dryRun`)

wpf500:
Looks great. I will hold back on approving this one while I release and test the first iteration of the newsletter groups logic because releasing this will start the cronjob and would be good to know the logic works first.

wpf500 (approved):

PR #639, opened 2026-06-26
feat(newsletter-integration): Catch error when contact is upserted to an invalid group

aasthas9:
# Catch upsert error and refresh group

[ticket](https://correctivdigital.openproject.com/projects/beabee/work_packages/2920/)

## Changes
1. Added error detail to upsert response interface
For group-related errors, the error detail says "invalid interest ID" as the interest ID does not exist anymore. This detail field was missing from the interface previously.
2. Catch invalid interest ID errors, trigger group refresh and cache update. Then retry the upsert. This retry is necessary so as to not leave the contact in "Pending" state.

**Note**: The error text is Mailchimp-specific. Will need to be changed/adapted for other providers as they are added.

### DIFF
--- a/packages/core/src/errors/CantUpdateNewsletterGroupsError.ts
+++ b/packages/core/src/errors/CantUpdateNewsletterGroupsError.ts
@@ -1,0 +1,22 @@
+import {
+  ApiErrorCode,
+  CantUpdateNewsletterGroupsErrorData,
+} from '@beabee/beabee-common';
+
+import { InternalServerError } from 'routing-controllers';
+
+export class CantUpdateNewsletterGroupsError
+  extends InternalServerError
+  implements CantUpdateNewsletterGroupsErrorData
+{
+  readonly httpCode = 400;
+  readonly code = ApiErrorCode.CANT_UPDATE_NEWSLETTER_GROUPS;
+
+  constructor(
+    readonly status: number,
+    readonly data: any
+  ) {
+    super("Can't update newsletter groups");
+    Object.setPrototypeOf(this, CantUpdateNewsletterGroupsError.prototype);
+  }
+}
--- a/packages/core/src/errors/CantUpdateNewsletterGroupsError.ts
+++ b/packages/core/src/errors/CantUpdateNewsletterGroupsError.ts
@@ -12,10 +12,7 @@ export class CantUpdateNewsletterGroupsError
   readonly httpCode = 400;
   readonly code = ApiErrorCode.CANT_UPDATE_NEWSLETTER_GROUPS;
 
-  constructor(
-    readonly status: number,
-    readonly data: any
-  ) {
+  constructor(readonly detail: string) {
     super("Can't update newsletter groups");
     Object.setPrototypeOf(this, CantUpdateNewsletterGroupsError.prototype);
   }
--- a/packages/core/src/providers/newsletter/MailchimpProvider.ts
+++ b/packages/core/src/providers/newsletter/MailchimpProvider.ts
@@ -101,7 +101,7 @@ export class MailchimpProvider implements NewsletterProvider {
       resp.status === 400 &&
       resp.data?.detail.toLowerCase().includes('invalid interest id')
     ) {
-      throw new CantUpdateNewsletterGroupsError(resp.status, resp.data);
+      throw new CantUpdateNewsletterGroupsError(resp.data.detail);
 
       // Try to put the user into pending state if they're in a compliance state
       // This can happen if they previously unsubscribed or were cleaned
--- a/packages/core/src/providers/newsletter/MailchimpProvider.ts
+++ b/packages/core/src/providers/newsletter/MailchimpProvider.ts
@@ -5,7 +5,10 @@ import {
 } from '@beabee/beabee-common';
 
 import config, { MailchimpNewsletterConfig } from '#config/config';
-import { CantUpdateNewsletterContactError } from '#errors/index';
+import {
+  CantUpdateNewsletterContactError,
+  CantUpdateNewsletterGroupsError,
+} from '#errors/index';
 import {
   createInstance,
   getMCMemberUrl,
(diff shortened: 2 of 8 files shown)


packages/core/src/services/NewsletterService.ts
```diff
@@ -87,13 +87,28 @@ class NewsletterService {
...
-        // newsletter status to None to prevent further updates
         if (err instanceof CantUpdateNewsletterContactError) {
-          log.warning(
-            `Newsletter upsert failed, setting status to none for contact ${contact.id}`,
-            err
-          );
+          // Tried to add contact to an invalid group. Refresh cached groups and retry upsert
+          if (
+            err.status === 400 &&
+            err.data.detail.toLowerCase().includes('invalid interest id')
```
wpf500: This checks for a Mailchimp specific error format in the `NewsletterService` which is always a good sign that the abstraction is at the wrong level.

The `MailchimpProvider` should handle the details of "what the Mailchimp error looks like", and my guess would be that it should then throw a new error type which the main service handles accordingly (e.g. `CantUpdateNewsletterContactGroupsError`, bit long bit consistent?).

wpf500 (changes requested):

packages/core/src/errors/CantUpdateNewsletterGroupsError.ts
```diff
@@ -0,0 +1,22 @@
...
+
+export class CantUpdateNewsletterGroupsError
+  extends InternalServerError
+  implements CantUpdateNewsletterGroupsErrorData
+{
+  readonly httpCode = 400;
+  readonly code = ApiErrorCode.CANT_UPDATE_NEWSLETTER_GROUPS;
+
+  constructor(
+    readonly status: number,
```
wpf500: I think these params are a bit too general to be useful and `any` is almost always a sign of something that can be improved :). Think: what do I want to know about this error? Ideally for this error I guess it would be _which_ group failed.

I think remove `status` and (it will always be 400 in the current implementation in any case, so is duplicate info with `httpCode`) and `data` (too general), and either:
* Add `readonly groupId: string` and parse the group ID out of the Mailchimp error (ideal but more complicated, might not be worth the effort)
* Add `readonly detail: string` and just give back the whole string.

packages/core/src/providers/newsletter/MailchimpProvider.ts
```diff
@@ -94,6 +97,14 @@ export class MailchimpProvider implements NewsletterProvider {
     if (resp.status === 200) {
       log.info('Updated member ' + contact.email);
       return mcMemberToNlContact(resp.data);
+    } else if (
+      resp.status === 400 &&
+      resp.data?.detail.toLowerCase().includes('invalid interest id')
+    ) {
+      log.info(
```
wpf500: This is very pedantic, but I think this log should live in the `NewsletterService` as _it_ is the place that triggers the refresh. Think: if the logic in `NewsletterService` changed then this log line would become out-of-date, but future developers wouldn't know as it's not next to the action itself. i.e. it assumes the behaviour of a different part of the codebase.

packages/core/src/services/NewsletterService.ts
```diff
@@ -87,12 +87,13 @@ class NewsletterService {
           { groups: newState.groups }
         );
       } catch (err) {
-        // The newsletter provider rejected the update, set this contact's
```
wpf500, line 91: This error still needs to be caught, was it intention that you removed it?

packages/core/src/services/NewsletterService.ts
```diff
@@ -87,12 +87,13 @@ class NewsletterService {
...
-        // newsletter status to None to prevent further updates
-        if (err instanceof CantUpdateNewsletterContactError) {
-          log.warning(
-            `Newsletter upsert failed, setting status to none for contact ${contact.id}`,
-            err
+        if (err instanceof CantUpdateNewsletterGroupsError) {
+          // Tried to add contact to an invalid group. Refresh cached groups and retry upsert
+
+          await this.refreshNewsletterGroups();
+          newState = await this.provider.upsertContact(
```
wpf500: I think I would make the method recover by calling itself rather than the provider method. This has the advantage that it will catch and handle other errors too (e.g. a `CantUpdateNewsletterContactError` occurs).

```ts
await this.upsertContact(contact, updates, opts);
return; // important: stop it proceeding further after it returns
```

wpf500 (changes requested):

packages/core/src/services/NewsletterService.ts
aasthas9: ah no. I misread the error type and thought the same error was caught twice.

packages/core/src/errors/CantUpdateNewsletterGroupsError.ts
aasthas9: I assumed `data: any` is used everywhere so that we're able to pass the full error in whatever format we receive it in, giving us all possible info for debugging?

packages/core/src/services/NewsletterService.ts
aasthas9: Won't calling the provider method do that too? `provider.upsertContact()` calls `upsertContactOrTryPending()` which would throw errors if something goes wrong again, and `NewsletterService` would handle them?

wpf500: Edit: just read your comment again and maybe this is actually more relevant: it wouldn't catch the errors any more because the call is in the `catch` block, it only catches errors throw in the `try` block. By calling the same method again you ensure errors are caught in a new `try` context.

Original: I think it wouldn't catch `CantUpdateNewsletterContactError`, which is the generic "something went wrong" that the newsletter provider can send. Also _theoretically_ the provider could rethrow the `CantUpdateNewsletterGroupsError`, obviously very unlikely but still nice to be able to handle it.

wpf500 (approved):

PR #642, opened 2026-06-29
feat(backend): Add health check cron job 

libby-correctiv:
- Added cron job to health check all integrations every 10 mins 
- Added Graylog logging for health checks

### Screenshots

If your change adds a new interface, or makes significant changes to an existing
one, add before and after screenshots.

## Checklist before merging

### DIFF
--- a/apps/backend/crontab
+++ b/apps/backend/crontab
@@ -4,7 +4,7 @@ PATH=/usr/local/bin:/usr/bin:/bin
 # Spread the jobs out over a few minutes to avoid overloading the server
 
 # Every 10 minutes
-*/10 * * * * su node -c 'export HOME=/home/node && cd /opt/apps/backend/ && yarn backend-cli health all' > /proc/1/fd/1 2>&1
+*/10 * * * * su node -c 'export HOME=/home/node && cd /opt/apps/backend/ && yarn backend-cli health all --notify' > /proc/1/fd/1 2>&1
 
 # Hourly cron jobs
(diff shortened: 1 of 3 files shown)


aasthas9 (commented):
Looks good to me, but I'll let @wpf500 review as well and merge the cron job PRs simultaneously.

wpf500 (changes requested):
Looks good, but one small improvement suggestion:

At the moment this will cause an error notification each time the command is run. This is useful for the cronjob but maybe not so much if we want to run this command manually.

I think it would be worth adding a CLI option (`--notify`?) which enables the `log.info`/`log.error` calls and then apply it to the cronjob.

e.g.
```sh
yarn backend-cli health all --notify # sends log messages (and thus triggers error notifications)
yarn backend-cli health all # no change from now, just prints to console
```

apps/backend/crontab
```diff
@@ -3,6 +3,9 @@ PATH=/usr/local/bin:/usr/bin:/bin
 
 # Spread the jobs out over a few minutes to avoid overloading the server
 
+# Every 10 minutes
+*/10 * * * * su node -c 'export HOME=/home/node && cd /opt/apps/backend/ && yarn backend-cli health all --notify' > /proc/1/fd/1 2>&1
```
wpf500: Sorry I didn't see this before, but can you add the `sleep $((RANDOM \% 200))` snippet to this command?

It randomises the start time of the cron job a bit so the cronjob doesn't trigger at exactly the same time for all instances. This is helpful to reduce server load and for avoiding getting rate limited by external providers (e.g. 50 requests to Mailchimp at the same time from the same IP could cause problems).

wpf500 (changes requested):

wpf500 (approved):

PR #643, opened 2026-07-01
feat: Clean up and notify after group refresh

aasthas9:
## Cleanup after group refresh and notify administrator if groups were deleted

[Ticket 1](https://correctivdigital.openproject.com/projects/beabee/work_packages/2807/) and [2](https://correctivdigital.openproject.com/projects/beabee/work_packages/2806/)

### Notification Email
Added a new default email template for automatic group deletion. This email is sent to the admin(s) if - following a group mismatch between app and newsletter provider - some groups are removed from the database. The email template is hidden from the UI to disable template editing.

🤖 AI was used for translating email templates from English to Dutch, German, Italian, Portugese and Russian

### Downstream group cleanup
When groups are found to be deleted, delete them from the tables "contact_profile", "callouts" and "content/join/setup".

🤖 AI was used for generating DB queries

image

## Checklist before merging


packages/core/src/data/email-templates.ts
```diff
@@ -221,6 +221,14 @@ export const adminEmailTemplates = {
       MEMBERNAME: params.contact.fullname,
     })
   ),
+  'deleted-group': withMergeFields(
+    ['GROUPS'],
+    (params: { groups: { id: string; label: string }[] }) => ({
+      GROUPS: params.groups
```
wpf500, line 227: I'm not actually sure if we can use HTML in these strings as I guess some email providers (like SendGrid) probably escape the contents. Do we do this anywhere else? Otherwise I think I would just convert this to a simple comma-separated list for simplicity.

packages/core/src/data/email-templates.ts
```diff
@@ -221,6 +221,14 @@ export const adminEmailTemplates = {
       MEMBERNAME: params.contact.fullname,
     })
   ),
+  'deleted-group': withMergeFields(
```
wpf500: Can you give this a more explicit name like `deleted-newsletter-groups` instead

packages/core/src/services/NewsletterService.ts
```diff
@@ -207,9 +210,70 @@ class NewsletterService {
...
+            ), '[]'::jsonb)`,
+          })
+          .where(`"newsletterGroups" ?| :removedIds`)
+          .setParameters({ removedIds })
+          .execute();
+
+        // 2. Remove from join flow (join/setup content)
+        log.info('Removing deleted groups from join/setup content');
+        await createQueryBuilder()
+          .update(Content)
```
wpf500: It's a bit of an anti-pattern to directly modify `Content` and `Callout` here as in theory they should only ever be directly accessed by their own services.

However in this case I think it is probably the best approach for now as to do this properly would introduce circular dependencies (`CalloutService` <-> `NewsletterService`). This is really where an events system (e.g. with BullMQ) is useful as then the newsletter service can send an event about group deletions and other services can listen for those events and handle the logic themselves.

wpf500 (changes requested):

wpf500 (approved):

PR #646, opened 2026-07-13
test(e2e-api-tests): Add API test for newsletter integration and refresh logic

aasthas9:
## Add API test for newsletter integrations

[Ticket](https://correctivdigital.openproject.com/projects/beabee/work_packages/2851/activity)

This PR adds a new e2e API test to primarily verify the cleanup logic after a newsletter group refresh.

### Test data update
1. Cached newsletter groups in test database
2. Added newsletter subscriptions to two contacts
3. Added newsletter group to join content

### New newsletter provider, only for testing
Added `TestProvider` to test newsletter integrations. Health check returns `HEALTHY` and the database NL group cache. Group refres returns a fixed list of groups (to simulate changes to newsletter groups)

### New API test
Tests the GET and POST endpoints and reconciliation logic.
1. GET should return provider == 'test' and a healthy status
2. GET should also return 3 groups from the test database
3. Refresh should detect that a group was deleted and delete it from options/newsletterGroups, contact_profile and content/join/setup.


.github/workflows/test-api.yml
```diff
@@ -17,6 +17,7 @@ jobs:
     runs-on: ubuntu-latest
     env:
       COMPOSE_PROJECT_NAME: beabee-api-tests-${{ github.run_id }}
+      BEABEE_NEWSLETTER_PROVIDER: test
```
wpf500: Should this not be in `.env.test`?

apps/backend/src/api/app.ts
```diff
@@ -105,7 +105,9 @@ initApp()
   })
   .then(() => {
     // Cache newsletter groups in DB
-    newsletterService.refreshNewsletterGroups();
+    if (config.newsletter.provider !== 'test') {
```
wpf500: Why is this needed?

.env.test.example
```diff
@@ -40,7 +40,7 @@ BEABEE_EMAIL_SETTINGS_AUTH_USER=dev
 BEABEE_EMAIL_SETTINGS_AUTH_PASS=dev
 
 # Test newsletter
-BEABEE_NEWSLETTER_PROVIDER=none
+BEABEE_NEWSLETTER_PROVIDER=test
```
wpf500, line 43: Ah you have it here, so I don't think it's needed in the action config?

apps/backend/src/api/app.ts
wpf500: Ah I think I see now. This is to stop the test provider doing an initial sync on startup?

This `if` statement strictly speaking means that this is no longer a "full" E2E test as the test logic differs to the real-world logic. Obviously in some cases this is unavoidable, but here I think we can do something cleaner :).

I think the `TestProvider` could:
1. Return semi-randomly generated groups and then the tests check if the before and after are as expected. Randomised test data can also be a bad idea but in this case I think it's okay
2. Or if you aren't keen on that, how about the `getGroups` method keeps an internal count of how many times it was called and returns something based on that?

wpf500 (changes requested):

apps/backend/src/api/app.ts
aasthas9: Yes for the why it's needed. The initial startup calls refresh, which causes the seeded cache to get overwritten before the tests can run.

1. The test needs to also check if contacts are unsubscribed and groups removed from join flow. This requires that we know and control which group is removed. For example, if Groups A, B and C were cached, and group A was in some contact subscriptions and the join flow, refresh *must* remove it or the test would fail. I think (semi)-randomising group return complicates the test logic unnecessarily.
2. Maintaining a count does not feel cleaner than the `if` to me. Is it?

.env.test.example
aasthas9: you're right. I'll take it out of the action

apps/backend/src/api/app.ts
wpf500: Ah yeah I see.

For me the count is cleaner because it keeps the test-specific logic inside the test provider, and doesn't modify the main application. As it currently is the test is still making specific assumptions about the application (e.g. that `refreshNewsletterGroups` isn't called before the test starts), I think it would be cleaner and less brittle if the assumptions it makes rely on something that also test-specific and not application-wide.

Let's talk about this later to see what the best way forwards is, as there are also other options (e.g. just leaving a comment, adding specific hooks into the `TestProvider`)

aasthas9: I've moved the group update to DevController. Also removed some unused imports.

wpf500 (approved):
Nice :). The dev controller change was definitely the way to go, maintains the separation between the test runner and the main app 👌

PR #648, opened 2026-07-13
fix: Display translated labels for callout segments

aasthas9:
## Display translated labels for callout segments

Callout segment labels were always displayed in English, even when the app language was something else. 

image

image

This was due to two reasons:
1. "All Responses" was added manually to the segment list and had no translations. This is now fixed by adding and using translations.
2. All other labels were obtained directly from the database and used as it is. This is fixed by comparing the segment label with the translated array of labels and returning the translation if it is available.

## Checklist before merging

### DIFF
--- a/apps/frontend/src/pages/admin/crowdnewsroom/view/[id]/responses/index.vue
+++ b/apps/frontend/src/pages/admin/crowdnewsroom/view/[id]/responses/index.vue
@@ -387,7 +387,7 @@ const {
   currentSegment,
 } = useSegmentManagement(
   `/admin/crowdnewsroom/view/${props.callout.slug}/responses`,
-  'All Responses',
+  t('calloutResponsesPage.allResponses'),
   listSegments,
   listTotalSegmentItems
 );
--- a/apps/frontend/src/pages/admin/crowdnewsroom/view/[id]/responses/index.vue
+++ b/apps/frontend/src/pages/admin/crowdnewsroom/view/[id]/responses/index.vue
@@ -251,6 +251,7 @@ import { addBreadcrumb } from '#store/breadcrumb';
 import { canAdmin } from '#store/currentUser';
 import { client } from '#utils/api';
 import { extractErrorText } from '#utils/api-error';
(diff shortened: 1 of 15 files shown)


apps/frontend/src/pages/admin/crowdnewsroom/view/[id]/responses/index.vue
```diff
@@ -420,11 +421,18 @@ async function updateSegment(
...
+  const segments = await client.callout.segments.list(
     props.callout.slug,
     { sort: 'order' },
     ['itemCount']
   );
+  return segments.map((segment) => ({
+    ...segment,
+    name:
+      buckets.value.find(
+        (bucket) => bucket.id.toLowerCase() === segment.name.toLowerCase()
```
Comment (wpf500) [inline apps/frontend/src/pages/admin/crowdnewsroom/view/[id]/responses/index.vue:433]: Ah I misunderstood the problem before. I would still say this is the right fix for now but it is a bit "dodgier" than I thought as it is actually caused by [a migration](https://github.com/beabee-communityrm/monorepo/blob/main/packages/core/src/migrations/1755526570700-AddDefaultCalloutResponseSegments.ts) that just always inserts segments with English names.

This is also a problem in the admin contacts list actually, as we also [create default segments](https://github.com/beabee-communityrm/monorepo/blob/b643af737b7d3c93f4b1f28ecd72470049d4dec1/packages/core/src/migrations/1616694307645-AddDefaultSegments.ts) for that too. I guess Svenja maybe just renames these for our clients or maybe everyone has English contact segments too?

We should come up with a better way of having default segments whose names are translated, but that's a bigger fix and out of scope for this PR.

wpf500 (approved):
Nice :)

apps/frontend/src/pages/admin/crowdnewsroom/view/[id]/responses/index.vue
Comment (aasthas9) [reply apps/frontend/src/pages/admin/crowdnewsroom/view/[id]/responses/index.vue:433]: Svenja only recently noticed this bug, so I assume everyone has contact segments in English and nobody noticed/flagged it.

PR #651, opened 2026-07-14
feat(frontend): Nuxt UI - Account page

libby-correctiv:
## Describe your changes

Migrates the member-facing Account page (contact info, delivery address, password, 2FA) from the legacy component stack (`AppApiForm`, `AppSlider`, `AppConfirmDialog`) to Nuxt UI (`UForm`, `UModal`, `UPinInput`, etc.).

- Adds a new `packages/vue/src/components/nuxt-ui/` set: `AppSectionCard`, `AppCodeInput`, `AppModalActions`, `AppCopyIconButton`, `AppFormSkeleton`, `AppIconBadge`, `AppStickySaveBar` — each with a story + doc comment.
- Adds a `useApiSubmit` composable for shared save/error/notify boilerplate.
- Rewrites `SetMFA.vue`'s enable flow from a single 3-slide wizard into two focused modals (enable, disable), removing the `AppSlider` step-validation bookkeeping.
- Adds a sticky "unsaved changes" save bar (`AppStickySaveBar`), teleported into the Dashboard layout's footer.
- Changed the form to use zod-based schema validation for `AccountForm.vue`/`ChangePassword.vue` instead of vuelidate.
- Adds `-nuxt`-suffixed locale keys alongside existing ones, since the legacy frontend still uses the old keys.

> Suggestion to rename `AppXxx` naming convention for the new Nuxt UI components. An alternative could be `UiXxx`?

### Screenshots
localhost_3000_profile_account
localhost_3000_profile_account (1)
localhost_3000_profile_account (2)
localhost_3000_profile_account (3)
localhost_3000_profile_account (4)

## Checklist before merging


packages/vue/src/assets/styles/fonts-fira-sans-condensed.css
```diff
@@ -1,4 +1,5 @@
 @import '@fontsource/fira-sans-condensed/300.css';
 @import '@fontsource/fira-sans-condensed/400.css';
+@import '@fontsource/fira-sans-condensed/500.css';
```
wpf500, line 3: Why was this font weight added? It would be good to check what this does to the page load weight as having lots of font weights can be expensive. Is it possible to remove one of the other font weights?

apps/frontend/src/components/pages/profile/account/ChangePassword.vue
```diff
@@ -1,79 +1,173 @@
...
+  showConfirmPassword.value = !showConfirmPassword.value;
+}
+
+function isValidPassword(value: string): boolean {
+  return !value || isPassword(value);
+}
+
+const formRef = useTemplateRef('formRef');
+
+// Deliberately no `v.nonEmpty(...)` "required" check here. The fields are initially
```
wpf500, line 114: Not sure I understand this comment. Aren't all required fields initially empty so would have the same behaviour?

apps/frontend/src/composables/useApiSubmit.ts
wpf500, line 1: Using a composable here is a great idea, much better than what we had before with `AppApiForm` I think 👍

apps/frontend/src/components/pages/profile/account/AccountForm.vue
```diff
@@ -0,0 +1,241 @@
...
+
+const dirty = ref(false);
+
+// UForm validates in response to DOM events (input/blur/change), not
+// reactively off `:state` — a silent `Object.assign` (in handleCancel below)
+// doesn't fire those events, so any already-shown errors would otherwise
+// stay stuck on screen. `formRef.value.clear()` explicitly wipes them.
+const formRef = useTemplateRef('formRef');
+
+watch(
```
wpf500: This is quite an expensive way to check if the data is dirty as each time anything changes all of `data` and `savedData` have to be stringified and it uses deep watchers.

In reality it doesn't matter as it's a small amount of data and computers are fast, but in terms of good practice it would be much better to use direct comparison checks (even if it is a bit uglier).

```ts
const dirty = computed(() =>
  data.email !== savedData.email &&
  // etc.
);
```

wpf500 (changes requested):

apps/frontend/src/components/pages/profile/account/ChangePassword.vue
libby-correctiv: The required fields in the rest of the page are pre-filled (in `AccountForm.vue`). In the `AccountForm.vue` it's handled in the usual way, where if the user deletes the value in a required input then blurs the input, an error immediately displays (because this input has gone from valid to invalid). 
But here it doesn't make sense because they may just be choosing not to fill out the form. I've updated the comment to make it clearer

packages/vue/src/assets/styles/fonts-fira-sans-condensed.css
libby-correctiv: It's Tailwind's `font-medium` and used a lot by Nuxt UI by default. (The most used weight I think). Nuxt UI uses 400, 500, 600 and 700 but doesn't use weight 300 by default, but we use it in the existing frontend. Could be removed in the clean up after the migration though

apps/frontend/src/components/pages/profile/account/AccountForm.vue
libby-correctiv: Ah makes sense :) Changed it

wpf500 (changes requested):
Review from UI testing:

**Needs fixing**

After saving it still says "You have unsaved changes". Presumably the new data should be copied into the saved data so the form is no longer marked as `dirty`?

Image

I think the toggle text for delivery opt-in should be on it's own line as it can be a large sentence. We could change this in the future but for now admins have defined it based on this so we should support it.

Image

--

**Optional**

Saved changes button wraps on mobile, not sure what the best multi-lingual compatible fix is but maybe no wrapping or just an icon?
Image

libby-correctiv:
@wpf500 
1. Save bar not disappearing after save - Oops, that was introduced from my changes from the last review. Fixed
2. now looks like this on narrow screens
Screenshot 2026-08-19 at 12 20 12

3. Saved changes button wraps on mobile - changed to just 'Save' :
Screenshot 2026-08-19 at 12 22 58

apps/frontend/src/components/pages/profile/account/AccountForm.vue
```diff
@@ -47,7 +47,9 @@
       <AppFormSkeleton v-if="loading" :rows="3" />
       <template v-else>
         <template v-if="accountContent?.showMailOptIn">
-          <div class="flex items-center justify-between gap-4">
+          <div
+            class="flex flex-col gap-4 sm:flex-row sm:items-start sm:justify-between"
```
wpf500: I believe long opt-in labels are always squeezed, not just on mobile. I think it should just always be stacked as clients have already provided strings with this layout in mind. For example, one client has set this text to "Check this box if you would like to receive the magazine by post".

The other option would be to stop using `accountContent.mailOptIn` altogether and just make it a simple toggle, but unfortunately for some clients `accountContent.mailText` and `mailOptIn` only make sense together, so I think it's better in this case just to always stack in.

wpf500 (changes requested):

wpf500 (approved):

PR #653, opened 2026-07-15
fix(browser-tests): Minor config improvements to reduce brittleness

aasthas9:
## Minor browser test config improvements

Browser tests have been failing randomly, sometimes passing when re-run and sometimes not. This PR:
- Removes explicit worker declaration from config: This is likely the primary culprit for tests breaking. Playwright picks a sane default automatically (number of logical cores / 2).
- Adds retries and marks tests as flaky in the report: Each failing test will be retried thrice. If a test passes on a retry, it will be flagged as flaky so we know something went wrong.
- Remove 8 second timeout when looking for payment history: After a page reload, the newly added payment row will either be there or not. It can't 'become' visible without a reload, so waiting 8 seconds is unnecessary. Defaulting to 5 seconds.

[Ticket](https://correctivdigital.openproject.com/projects/beabee/work_packages/2970/)

### DIFF
--- a/apps/browser-tests/src/tests/one-time-contribution.spec.ts
+++ b/apps/browser-tests/src/tests/one-time-contribution.spec.ts
@@ -55,9 +55,8 @@ test("One-time contribution after login", async ({ page }) => {
   for (let attempt = 1; attempt <= 5; attempt++) {
     await page.reload();
     try {
-      await expect(targetRow, "Payment date and amount visible").toBeVisible({
-        timeout: 8000,
-      });
+      console.log("Reloading after payment. Reload #", attempt);
+      await expect(targetRow, "Payment date and amount visible").toBeVisible();
       break;
     } catch (e) {
       if (attempt === 5) throw e;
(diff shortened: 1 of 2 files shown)


apps/browser-tests/src/tests/one-time-contribution.spec.ts
```diff
@@ -55,9 +55,8 @@ test("One-time contribution after login", async ({ page }) => {
   for (let attempt = 1; attempt <= 5; attempt++) {
     await page.reload();
     try {
-      await expect(targetRow, "Payment date and amount visible").toBeVisible({
-        timeout: 8000,
-      });
+      console.log("Reloading after payment. Reload #", attempt);
+      await expect(targetRow, "Payment date and amount visible").toBeVisible();
```
wpf500, line 59: I guess the timeout was however still useful for delaying reloading the page again? Would it be worth just having a quick 1s sleep or something in the `catch` clause if it will reload?

aasthas9: There's a 5 second timeout for `expect` by default, so the page wouldn't reload immediately.

wpf500 (approved):

PR #655, opened 2026-07-16
feat(uploads): use streaming for file handling instead of buffers

wpf500:
Image and document uploads/downloads previously buffered the entire file in memory at every hop: nginx spooled the full request body, multer held it in `req.file.buffer`, and the services uploaded via a single `PutObjectCommand({ Body: <Buffer> })`. Downloads were worse — `GET /images/:id` downloaded the full original from S3 **twice** per request (once inside `getImageMetadata` just to read width/height, once for the response body).

This PR streams uploads and downloads end-to-end for both endpoints, and backfills S3 metadata for images uploaded before the change.

## What changed

- **`DocumentService.uploadDocument`** streams straight through — the `%PDF` magic-byte check peeks 4 bytes off the stream and pushes them back (`peekStreamBytes`) rather than needing the whole buffer.
- **`ImageService.uploadImage`** pipes the request stream into sharp, which probes dimensions/format before the encoded output streams straight to S3 (no more full-buffer `.toBuffer()` round trip). SVGs are still fully buffered (SVGO needs the whole document), and SVG content under a non-`.svg` extension is now rejected with 400 instead of being silently detected and converted — a small, deliberate behavior change since the raw bytes aren't recoverable once sharp has consumed the stream.
- **`getImageMetadata` is now a single `HeadObject` call.** Width/height are written to S3 object metadata at upload time and read back directly, eliminating the double full-image download on every `GET /images/:id`. Legacy images without stored dimensions report `0` (nothing currently consumes those fields).
- **Downloads stream from S3 straight to the response** (`pipeline(stream, res)`) in both controllers instead of buffering into a `Buffer` return value.
- **New CLI command**: `yarn backend-cli image backfill-dimensions [--dryRun]` retroactively writes width/height metadata for images uploaded before this change, via a server-side `CopyObjectCommand` self-copy (no byte re-upload). Idempotent — already-backfilled images are skipped.
- **New e2e coverage**: 7 invalid-upload cases (empty file, junk bytes, truncated image, mislabeled SVG, unsupported format, oversized file, plain text) added to `upload.test.ts`, generated in memory rather than as committed fixtures.

## Test plan

- [x] `yarn format && yarn check && yarn build` all pass
- [x] `yarn test` (unit suites) passes
- [x] nginx config syntax-checked in an `nginx:alpine` container
- [x] e2e upload tests (`apps/e2e-api-tests`)
- [ ] `image backfill-dimensions` — not yet run against a real MinIO instance (blocked on the same stack issue)


packages/core/src/services/ImageService.ts
```diff
@@ -180,16 +186,29 @@ export class ImageService {
...
+          ? getMimetypeFromExtension(metadata.format)
+          : getMimetypeFromExtension(extension);
 
-      // Handle SVG optimization
-      if (metadata.format === 'svg' && outputFormat === 'original') {
-        // Convert buffer to string for SVGO
-        const svgString = imageData.toString('utf8');
+      // Prepare metadata for S3, storing the dimensions so they can be read
+      // later without downloading the image
+      const s3Metadata: S3Metadata = {
```
aasthas9, line 206: If the image is rotated later (line 239), this metadata would be incorrect, wouldn't it?

wpf500: Nice catch! Fixed with 42e391e358a339d609ec1d6ba098e8cc36bf1154

aasthas9 (commented):
If the Docker stack does not boot, merging this PR would break subsequent workflows on `main` as well as any future branches. Is it possible to fix the import error as well?

wpf500:
Opps sorry @aasthas9 that was AI junk, I've run the E2E tests and [they passed](https://github.com/beabee-communityrm/monorepo/actions/runs/29730162357).

aasthas9 (dismissed):

aasthas9 (approved):

PR #658, opened 2026-07-21
feat(core): Salesforce newsletter provider (stage 1, outbound)

wpf500:
## Context

CORRECTIV wants Salesforce as an alternative newsletter provider so member/newsletter changes in beabee are pushed into the Salesforce (NPSP) org, where newsletter preferences live as boolean fields directly on the `Contact` object.

This is **stage 1: the outbound provider only** (beabee → Salesforce). It mirrors the existing `MailchimpProvider`, plugging into the same `NewsletterProvider` abstraction.

## What's included

- **`SalesforceNewsletterConfig`** (`config.ts`) — config-driven field mapping (`subscriptionField`, `groupFieldMap`, `mergeFieldMap`, active-member/user fields) so the same code works across org schemas; new `salesforce` provider option + env branch.
- **`SalesforceNewsletterIntegrationData`** added to the `NewsletterIntegrationData` union in `@beabee/beabee-common`.
- **`lib/salesforce.ts`** — REST client with OAuth2 password-grant auth (token caching + one 401 retry), `findContactByEmail` (SOQL, single-quote escaped), `getContactById`, and config-driven mappers both directions.
- **`SalesforceProvider.ts`** — implements all 7 `NewsletterProvider` methods. Matches Contacts by email; **creates one if missing** on upsert. `permanentlyDeleteContact` anonymises PII rather than hard-deleting (Salesforce blocks delete when linked Opportunity/Recurring Donation records exist).
- Wired into `NewsletterService.createProvider()`. `ContactsService` already triggers newsletter sync on every material contact change, so no service-layer changes were needed.
- `.env.example` documents all new `BEABEE_NEWSLETTER_SETTINGS_*` vars.

## Design decisions

| Decision | Choice |
|---|---|
| Scope | Outbound only — inbound (CDC/polling) deferred to stage 2 |
| Missing contact on upsert | Create it (mirrors Mailchimp PUT-upsert) |
| Bulk sync | Deferred — falls through to `NoneBulkProvider` |
| Auth | OAuth2 password grant (matches existing reference integrations; note SF retires it Summer 2027 → JWT migration is a future task) |

## Out of scope
Payment sync, inbound CDC/polling, bulk provider, storing the Salesforce Contact ID on the beabee Contact model.

## Before it can be tested live
Requires org details not yet available: real field API names (which is the master subscription field), connected-app credentials, the minimum required fields to create a Contact (likely `LastName`), and a pinned REST API version. The config-driven design lets the code ship now and the field map be filled in later.

## Verification
- `@beabee/core` `build` and `check` (tsc + prettier + class-index) pass — exit 0.
- No inbound/webhook verification this stage (deferred).

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -128,9 +128,10 @@ BEABEE_NEWSLETTER_PROVIDER=none
 # BEABEE_NEWSLETTER_SETTINGS_WEBHOOKSECRET=
 
 # Salesforce settings (when provider=salesforce)
-# BEABEE_NEWSLETTER_SETTINGS_AUTHURL=https://login.salesforce.com/services/oauth2/token
-# BEABEE_NEWSLETTER_SETTINGS_USERNAME=api-user@example.com
-# BEABEE_NEWSLETTER_SETTINGS_PASSWORD=your_password
+# Auth uses the OAuth2 client-credentials flow. The connected app must have
+# "Enable Client Credentials Flow" set and a Run-As user configured; that user's
+# field permissions govern what can be read/written. Use the My Domain URL.
+# BEABEE_NEWSLETTER_SETTINGS_AUTHURL=https://correctiv.my.salesforce.com/services/oauth2/token
 # BEABEE_NEWSLETTER_SETTINGS_CLIENTID=connected_app_client_id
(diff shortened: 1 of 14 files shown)


wpf500:
Closing this PR until the scope of the Salesforce integration is clearer. The branch will remain.

PR #666, opened 2026-08-04
feat(frontend): Nuxt UI - sidebar dark mode styling, color improvements, footer layout tweaks

libby-correctiv:
## Describe your changes

- Remove some hardcoded background colouring that was making dark mode look weird, changed to `bg-default` which changes automatically based on user's dark/light mode setting
- Small styling tweak to center footer items
- Colour improvements - we were previously computing colour shades manually in `setNuxtShades` from the user's chosen  colours using RGB-mixed shades. This was resulting in washed-out colour shades compared to Tailwind's OKLCH-based colouring. Added a tiny package which computes Tailwind-style shades. Colour shades are closer now but still less vibrant than Tailwind colours - for this reason I removed the `nuxt-neutral` colour for now and set it to `gray` (a built in Tailwind colour). Can come back to this later if we decide we want user-defined neutral colour
- Also add Nuxt UI auto-generated type declarations to `.prettierIgnore` (flagged in yarn check)

### Screenshots

localhost_3000_ (2)
Looks weird because its just the sidebar :-)

## Checklist before merging

### DIFF
--- a/packages/vue/package.json
+++ b/packages/vue/package.json
@@ -22,6 +22,7 @@
     "generate:index": "yarn workspace @beabee/dev-cli start generate-index --extension ts --baseDir $(pwd) -p ./src/lib ./src/store ./src/types ./src/utils"
   },
   "dependencies": {
+    "@bobthered/tailwindcss-palette-generator": "^4.0.0",
     "@fontsource/fira-sans": "^5.2.7",
     "@fontsource/fira-sans-condensed": "^5.2.7",
     "@fontsource/libre-franklin": "^5.2.8",
--- a/apps/frontend/src/components/TheFooter.vue
+++ b/apps/frontend/src/components/TheFooter.vue
@@ -1,7 +1,7 @@
 <template>
   <UFooter
     :ui="{
-      root: 'border-t border-default',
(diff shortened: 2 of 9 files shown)


packages/vue/package.json
```diff
@@ -22,6 +22,7 @@
     "generate:index": "yarn workspace @beabee/dev-cli start generate-index --extension ts --baseDir $(pwd) -p ./src/lib ./src/store ./src/types ./src/utils"
   },
   "dependencies": {
+    "@bobthered/tailwindcss-palette-generator": "^4.0.0",
```
aasthas9, line 25: @wpf500 could you have a look at the new package dependency?

aasthas9 (commented):
I'm not entirely sure about how we assess and add dependencies, so I'll hold back on approving this until Will comments.

packages/vue/package.json
wpf500: Always good to question new dependencies for sure. In this case the library has been stable for two years and has none of its own dependencies, which significantly lowers the security risks.

I think it's okay 👍

wpf500 (approved):

PR #668, opened 2026-08-11
feat: Display sign-up source, referral and campaign information on contact account page

aasthas9:
## Display sign-up source, referral and campaign information on contact account page

Tickets [1](https://correctivdigital.openproject.com/projects/beabee/work_packages/3071), [2](https://correctivdigital.openproject.com/projects/beabee/work_packages/3073), [3](https://correctivdigital.openproject.com/projects/beabee/work_packages/3074)

### Changes
1. (frontend) New "Origin" section on contact page (visible to admins only). Contains three fields - sign up source, referrer and campaign
2. New `GetContactWith.Origin` option: Requesting contact info with `?with=origin` returns contact with the origin information. The origin information is obtained from the metadata stored for that contact's `contact.created` event
3. On a new sign-up from the join flow, the URL is parsed for UTM parameters. `utm_source`, `utm_medium` and `utm_campaign` provide the sign-up source, referrer and campaign, respectively. Added a new "origin" column to the `signup_flow` table to store this info during signup. Once sign-up succeeds and the contact is added, signup flow passes the information to ContactsService, which will add an event via ActivityService and store the info as metadata
4. On a new sign-up via a callout, the sign-up source is set as 'callout' manually 

### Screenshots

image

Contacts added via callouts show callout slug as campaign name

image

Contacts with no stored information display empty fields

## Checklist before merging


packages/core/src/services/SignupService.ts
```diff
@@ -252,13 +254,16 @@ class SignupService {
...
     } else {
-      contact = await ContactsService.createContact(partialContact, {
-        newsletterStatus: OptionsService.getText('newsletter-default-status'),
-        ...(completedFlow?.data.billingAddress &&
-          OptionsService.getBool('show-mail-opt-in') && {
-            deliveryAddress: completedFlow.data.billingAddress,
-          }),
-      });
+      contact = await ContactsService.createContact(
+        { ...partialContact, origin: signupFlow.source },
```
wpf500, line 258: I think the pattern of passing `origin` in as part of the `partialContact`, only to then destructure out of `partialContact` within `createContact` is a bit strange. It might be cleaner as a third parameter to `createContact`?

packages/core/src/models/SignupFlow.ts
```diff
@@ -48,4 +50,7 @@ export class SignupFlow {
 
   @Column({ default: false })
   processing!: boolean;
+
+  @Column({ type: 'jsonb', nullable: true })
+  source!: ContactOriginData | null;
```
wpf500, line 55: For naming consistency I would call this field `origin`

packages/core/src/models/ActivityEvent.ts
```diff
@@ -30,5 +34,5 @@ export class ActivityEvent {
   eventType!: ActivityEventType;
 
   @Column({ type: 'jsonb', nullable: true })
-  metadata!: Record<string, string> | null;
+  metadata!: ContactOriginData | Record<string, string> | null;
```
wpf500, line 37: This is fine for now but we can actually type this field much more strictly with generics which will be a nicer way to do this when more events are used. You could establish this pattern here or do it in a future PR

Tip: you are aiming for something like this I think
```ts
eventType: T;
metadata: ActivityEventMetadata<T> | null;
```

apps/frontend/src/pages/join/index.vue
```diff
@@ -51,6 +52,20 @@ const router = useRouter();
 
 const stripeClientSecret = ref('');
 
+// UTM parameters parsed form URL
+const origin = ref<ContactOriginData>({
```
wpf500, line 56: I believe this could be a `computed` as `route` is a reactive variable. This also catches an edge case where the query params change once the page has been loaded. In reality I guess not really likely, but it's cleaner to use `computed` wherever possible to avoid unnecessary duplicate state.

```ts
const origin = computed<ContactOriginData>(() => {
  return {...};
});
```

apps/frontend-old/src/pages/join/index.vue
```diff
@@ -51,6 +52,20 @@ const router = useRouter();
 
 const stripeClientSecret = ref('');
 
+// UTM parameters parsed form URL
+const origin = ref<ContactOriginData>({
```
wpf500, line 56: Same change here

apps/frontend-old/src/pages/join/index.vue
```diff
@@ -139,6 +155,12 @@ async function submitStep1() {
 onBeforeMount(async () => {
   stripeClientSecret.value = '';
 
+  origin.value = {
+    source: getUrlRouteAsString(route.query.utm_source),
+    referrer: getUrlRouteAsString(route.query.utm_medium),
```
wpf500, line 160: I worry a bit about the naming convention here. `referrer` seems like the wrong name if it's being mapped to `utm_medium`, maybe it should just be `medium` everywhere too for consistency? Happy to discuss options

wpf500 (changes requested):

apps/frontend-old/src/pages/join/index.vue
aasthas9: I used referrer for consistency with the frontend copy. But it _is_ confusing, I agree. I'll change it to medium with the comment that it is displayed under 'referrer'?

packages/core/src/models/ActivityEvent.ts
aasthas9: Good point. I'll add this typing on the branch containing a bulk of the activity feed changes rather than here.

wpf500 (approved):

PR #669, opened 2026-08-12
feat(activity-feed): Add table for activity feed events

aasthas9:
## (feat) Add table for activity feed events

Ticket [1](https://correctivdigital.openproject.com/projects/beabee/work_packages/2999/)

This PR adds a new database migration that creates the table 'activity_events' . Also includes some index file changes from `generate:index`


wpf500 (changes requested):
Can you move the model definition into this PR too? Otherwise the migration and database state will be out of sync, probably not a problem in reality but it would mean if someone else generated a migration before your next PR is ready it would include a `DROP TABLE` statement for the activity feed!

packages/core/src/models/ActivityEvent.ts
```diff
@@ -0,0 +1,32 @@
...
+@Index(['targetId', 'createdAt'])
+export class ActivityEvent {
+  @PrimaryGeneratedColumn('uuid')
+  id!: string;
+
+  @Column({ type: 'uuid', nullable: true })
+  targetId!: string | null;
+
+  @Column({ type: String })
+  contactId!: string;
```
wpf500, line 19: I think this should be `actorType` to match the migration?

wpf500 (changes requested):

packages/core/src/models/ActivityEvent.ts
aasthas9: The two branches are a bit out-of-sync because the model changed over a couple of commits. I originally split the PRs for database and activity service thinking that the service implementation would take a while.
But now that we're merging both at the same time, it's better to do it as a single PR.

aasthas9:
Closing this PR in favor of #673  which contains changes for the activity feed table as well as the service.

PR #670, opened 2026-08-12
feat(activity-feed): Create service to manage activity feed

aasthas9:
# Create service to add and fetch activity feed events

This PR introduces preliminary changes for the activity feed. Events are not yet added to, or fetched from, the database.

## Changes
1. Added `ActivityService` to manage activity feed events
2. Added enums for actor type and event type
3. Cleanup: removed unused imports


packages/core/src/services/ActivityService.ts
```diff
@@ -0,0 +1,38 @@
...
+  /**
+   * Add new event to events table
+   * @param event Event to add
+   */
+  async addEvent(
+    event: Pick<
+      ActivityEvent,
+      'targetId' | 'actorId' | 'actorType' | 'eventType' | 'metadata'
+    >
+  ): Promise<InsertResult> {
```
wpf500, line 16: I think it would be more consistent with other services to return `void` or the the saved `ActivityEvent` here, unless you have a specific reason for exposing the `InsertResult`?

You can use `getRepository().save` instead of `insert` to return the saved entity (with the generated ID set etc.).

packages/core/src/models/ActivityEvent.ts
```diff
@@ -18,15 +20,18 @@ export class ActivityEvent {
   @Column({ type: String })
   contactId!: string;
```
wpf500, line 21: I guess this is just a cherry-picking style problem between this and the previous PR, but I don't think `contactId` is needed here in any case?

I think once #669 is right this diff can just be updating the types for `actorType` and `eventType`

wpf500 (changes requested):

aasthas9:
Closing this PR in favor of #673, which contains changes for the activity feed table as well as the service.

PR #673, opened 2026-08-13
feat(activity-feed): Add database table for activity feed events, implement service

aasthas9:
## Add DB table and service to store and manage activity feed events

This PR introduces preliminary changes for the activity feed. Events are not yet added to, or fetched from, the database. Also includes some index file changes from `generate:index`

## Changes
1. Add model for activity feed event, add new database migration that creates the table 'activity_events' 
2. Added `ActivityService` to manage activity feed events
3. Added enums for actor type and event type
4. Cleanup: removed unused imports


packages/common/src/data/activity-actor.ts
wpf500, line 1: Very picky but can this file be called `activity-actor.ts` (without the `-event`) as the types defined within it are `ActivityActor*`, not `ActivityEventActor*`

packages/core/src/services/ActivityService.ts
```diff
@@ -0,0 +1,39 @@
...
+   * Add new event to events table
+   * @param event Event to add
+   */
+  async addEvent(
+    event: Pick<
+      ActivityEvent,
+      'targetId' | 'actorId' | 'actorType' | 'eventType' | 'metadata'
+    >
+  ): Promise<void> {
+    await getRepository(ActivityEvent).insert(event);
```
wpf500, line 18: Copied from other PR:

I think it would be more consistent with other services to return `void` or the the saved `ActivityEvent` here, unless you have a specific reason for exposing the `InsertResult`? We don't do this anywhere else.

You can use `getRepository().save` instead of `insert` to return the saved entity (with the generated ID set etc.).

wpf500 (changes requested):

packages/core/src/services/ActivityService.ts
aasthas9: It returns `void` now. Is it still preferable to use `save()` instead of `insert()`?

packages/common/src/data/activity-actor.ts
aasthas9: done

packages/core/src/services/ActivityService.ts
wpf500: `insert` is fine (and more efficient) if you don't need the saved entity.

wpf500 (approved):

PR #677, opened 2026-08-17
fix: add rule to SQL conversion tests and fix filter handler bugs

wpf500:
Adds a unit test suite for the rule group → SQL conversion (`convertRulesToWhereClause`) and fixes two bugs in the contact filter handlers that the tests surfaced.

## Tests (`packages/core/src/utils/rules.test.ts`)

- Documents the exact SQL produced for every filter type and operator
- Replays a real contact search request against the real `contactFilterHandlers` (callout `hasAnswered` + `responses.createdAt` rules).
- Also fills `validateRuleGroup`/`flattenRules` gaps in the `@beabee/beabee-common` tests.

## Fixes

1. **Callout response subqueries leaked the outer query's alias** (`calloutsFilterHandler`): the outer field prefix was forwarded into the response subquery, which only worked when the outer alias happened to be `item`. `batchSelect` uses `entity`, producing invalid SQL like `entity.createdAt` — this broke PATCH `/1.0/contact` with callout response rules (`column entity.createdat does not exist`). The subquery now uses its own alias, renamed to `response` for clarity.
2. **`activePermission` interpolated the rule value into SQL** (`cr.type = '${roleType}'`): not exploitable since validation restricts the value to the role enum, but it was the only place a rule value reached the query as text instead of a bound parameter. Now bound as `:roleType`.

Both fixes have regression tests pinning the corrected SQL.

### DIFF
--- a/packages/core/src/filter-handlers/contact.filter-handlers.ts
+++ b/packages/core/src/filter-handlers/contact.filter-handlers.ts
@@ -139,7 +139,7 @@ const activePermission: FilterHandler = (qb, args) => {
     .subQuery()
     .select(`cr.contactId`)
     .from(ContactRole, 'cr')
-    .where(`cr.type = '${roleType}'`)
+    .where(args.addParamSuffix('cr.type = :roleType'))
     .andWhere(`cr.dateAdded <= :now`)
     .andWhere(
       new Brackets((qb) => {
@@ -152,6 +152,8 @@ const activePermission: FilterHandler = (qb, args) => {
   } else {
     qb.where(`${args.fieldPrefix}id NOT IN ${subQb.getQuery()}`);
   }
+
+  return { roleType };
 };
 
 /**
--- a/packages/core/src/filter-handlers/contact.filter-handlers.ts
+++ b/packages/core/src/filter-handlers/contact.filter-handlers.ts
@@ -176,19 +176,24 @@ const calloutsFilterHandler: FilterHandler = (qb, args) => {
     case 'responses': {
       const subQb = createQueryBuilder()
         .subQuery()
-        .select('item.contactId')
-        .from(CalloutResponse, 'item');
+        .select('response.contactId')
+        .from(CalloutResponse, 'response');
 
       const responseField = restFields.join('.');
       const filterHandler = getFilterHandler(
         calloutResponseFilterHandlers,
         responseField
       );
-      params = filterHandler(subQb, { ...args, field: responseField });
+      // Pass the subquery's alias as the field prefix, not the outer query's
+      params = filterHandler(subQb, {
+        ...args,
+        fieldPrefix: 'response.',
+        field: responseField,
+      });
 
       subQb
-        .andWhere(args.addParamSuffix('item.calloutId = :calloutId'))
-        .andWhere('item.contactId IS NOT NULL');
+        .andWhere(args.addParamSuffix('response.calloutId = :calloutId'))
+        .andWhere('response.contactId IS NOT NULL');
 
       qb.where(`${args.fieldPrefix}id IN ${subQb.getQuery()}`);
       break;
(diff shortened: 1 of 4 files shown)


packages/core/src/filter-handlers/contact.filter-handlers.ts
```diff
@@ -176,19 +178,24 @@ const calloutsFilterHandler: FilterHandler = (qb, args) => {
     case 'responses': {
       const subQb = createQueryBuilder()
         .subQuery()
-        .select('item.contactId')
-        .from(CalloutResponse, 'item');
```
wpf500, line 180: This here is the key bug fix. The contact callout filter (`contactFilterHandlers.callouts.*`) creates a subquery which hard-coded the `CalloutResponse` alias to `item`...

packages/core/src/filter-handlers/contact.filter-handlers.ts
```diff
@@ -176,19 +178,24 @@ const calloutsFilterHandler: FilterHandler = (qb, args) => {
...
+        .select('response.contactId')
+        .from(CalloutResponse, 'response');
 
       const responseField = restFields.join('.');
       const filterHandler = getFilterHandler(
         calloutResponseFilterHandlers,
         responseField
       );
-      params = filterHandler(subQb, { ...args, field: responseField });
```
wpf500, line 187: ... but then defers to the callout response filter handler with `...args` which contains a `fieldPrefix` (in the case of the bug set to `"entity"`). This means the deferred filter handler references `"entity"` instead of the callout response.

packages/core/src/filter-handlers/contact.filter-handlers.ts
```diff
@@ -176,19 +178,24 @@ const calloutsFilterHandler: FilterHandler = (qb, args) => {
     case 'responses': {
       const subQb = createQueryBuilder()
         .subQuery()
-        .select('item.contactId')
-        .from(CalloutResponse, 'item');
+        .select('response.contactId')
```
wpf500, line 181: This could have stayed the same but I've renamed it for clarity as this is a common pattern across subqueries in filter handlers...

packages/core/src/filter-handlers/contact.filter-handlers.ts
```diff
@@ -176,19 +178,24 @@ const calloutsFilterHandler: FilterHandler = (qb, args) => {
...
       const responseField = restFields.join('.');
       const filterHandler = getFilterHandler(
         calloutResponseFilterHandlers,
         responseField
       );
-      params = filterHandler(subQb, { ...args, field: responseField });
+      // Pass the subquery's alias as the field prefix, not the outer query's
+      params = filterHandler(subQb, {
+        ...args,
+        fieldPrefix: 'response.',
```
wpf500, line 192: This is the real fix: the `fieldPrefix` needs to be overridden to reference the correct table. As per above this could have just been `fieldPrefix: 'item.'` but it uses the new cleaner table alias

packages/core/src/utils/rules.ts
```diff
@@ -385,8 +384,7 @@ export async function batchSelect<
   filters: Filters<Field>,
   ruleGroup: RuleGroup,
   contact?: Contact,
-  filterHandlers?: FilterHandlers<Field>,
-  queryCallback?: (qb: SelectQueryBuilder<Entity>, fieldPrefix: string) => void
```
wpf500, line 389: I just removed this as it wasn't being used and was confusing me with it's reference to `'entity.'` below

packages/core/src/filter-handlers/contact.filter-handlers.ts
```diff
@@ -139,7 +139,7 @@ const activePermission: FilterHandler = (qb, args) => {
     .subQuery()
     .select(`cr.contactId`)
     .from(ContactRole, 'cr')
-    .where(`cr.type = '${roleType}'`)
+    .where(args.addParamSuffix('cr.type = :roleType'))
```
wpf500, line 142: This change is unrelated to the bug but fixes a potential SQL injection. In reality there is no current risk as `roleType` is validated against a whitelist of strings (`"member"`, `"admin"`, etc.`), but it's still cleaner to add it as a parameter to protect against future regressions

aasthas9 (dismissed):

wpf500:
@aasthas9 I've added some documentation that hopefully makes the logic a bit clearer, can you re-review?

aasthas9 (approved):

PR #679, opened 2026-08-24
feat(frontend): Member subscription self management

libby-correctiv:
Adds a "Subscriptions" tab to the account page where members can view and unsubscribe from individual newsletter groups. Also makes group updates safer against a narrow race condition where a group added on Mailchimp's side could get overwritten if the sync webhook is down or hasn't caught up yet.

### Changes
**Frontend**
- New tabbed layout on the account page (Contact Information / Security / Subscriptions)
- New `AccountNewsletterSubscriptions.vue` component and locale strings
- New `GET`/`DELETE /contact/:id/newsletter-groups[/:groupId]` endpoints + client methods
- Added a small info message reading "Group names will be displayed to members in the subscription management tab of the Account page" to the Integrations tab in the Settings page

**Backend**
- Replaced the boolean `mergeGroups` option with `NewsletterGroupChange` (`'add' | 'remove' | 'replace'`) - `add`/`remove` now scope the Mailchimp `interests` payload to only the group(s) being changed, instead of reasserting the full list, to avoid overwriting a group Mailchimp already knows about that beabee's local DB hasn't caught up on yet
- Updates without groups now omit `interests` from the Mailchimp payload entirely, instead of reaffirming the full list on unrelated updates
- Add a retry limit to `NewsletterService.upsertContact`'s invalid-group recovery - previously retried indefinitely if a group ID was permanently invalid (not just stale), silently hanging instead of surfacing an error

**Testing** 
- new `mailchimp.test.ts` covering the payload shape for add/remove/replace; `newsletter.test.ts` updated. `yarn test` passing.

### Screenshots

Screenshot 2026-08-19 at 11 16 07
Screenshot 2026-08-19 at 11 16 18

## Checklist before merging

## Test scenarios

- [x] Someone joins via the join form, doesn't subscribe to any groups
- [x] Someone joins via the join form, no groups show, just subscribe tickbox
- [x] Someone joins via the join form, does subscribe to groups
- [x] Someone with an existing account answers a callout, updates groups
- [x] Someone with no account answers a callout, sets groups
- [x] Someone added via Mailchimp UI, groups are synced
- [x] Someone added/removed from Mailchimp UI


apps/frontend/src/pages/profile/account.vue
```diff
@@ -28,13 +39,33 @@ meta:
...
+import AccountNewsletterSubscriptions from '../../components/pages/profile/account/AccountNewsletterSubscriptions.vue';
 import ChangePassword from '../../components/pages/profile/account/ChangePassword.vue';
 import SetMFA from '../../components/pages/profile/account/SetMFA.vue';
 import { addBreadcrumb } from '../../store/breadcrumb';
 import { routeIcons, routeLabels } from '../../utils/route-nav';
+import type { TabsItem } from '@nuxt/ui';
+
+const items = ref<TabsItem[]>([
+  {
+    label: 'Contact Information',
```
wpf500: Need to be internationalised (and other labels below)

apps/frontend/src/type/newsletter-group.ts
wpf500, line 1: Could this be in `@beabee/beabee-common` and used in the backend too?

packages/core/src/services/NewsletterService.ts
```diff
@@ -82,11 +83,36 @@ class NewsletterService {
...
+    return this.upsertContactWithRetry(contact, updates, opts, 0);
+  }
+
+  /**
+   * Same as `upsertContact`, but tracks how many times an invalid-group
+   * failure has already been retried, so a group ID that's permanently
+   * invalid (not just stale in our cache) fails loudly instead of retrying
+   * forever.
+   */
+  private async upsertContactWithRetry(
```
wpf500: Why was the retry logic introduced?

packages/core/src/services/CalloutsService.ts
```diff
@@ -282,7 +282,7 @@ class CalloutsService {
           newsletterStatus: NewsletterStatus.Pending,
           newsletterGroups: newsletter.groups,
         },
-        { mergeGroups: true }
+        { newsletterGroupChange: 'add' }
```
wpf500, line 285: Has this been tested?

apps/backend/src/api/controllers/ContactController.ts
```diff
@@ -500,4 +502,32 @@ export class ContactController {
...
+   * @param target The target contact
+   * @param groupId The newsletter group ID
+   */
+  @OnUndefined(204)
+  @Delete('/:id/newsletter-groups/:groupId')
+  async unsubscribeNewsletterGroup(
+    @TargetUser() target: Contact,
+    @Params() { groupId }: { groupId: string }
+  ): Promise<void> {
+    await ContactsService.unsubscribeFromNewsletterGroup(target, groupId);
```
wpf500: `getNewsletterGroups` uses the `NewsletterService` but `unsubscribeNewsletterGroup` uses `ContactsService`...
 
This sort of non-symmetry can often be a sign that there could be a better level of abstraction, question to ask: why is one service handling one part of the logic but another handling another part?

packages/core/src/utils/newsletter.ts
wpf500, line 1: This touches quite a core method in newsletter syncing, have all paths been tested?

wpf500 (commented):
This is a really great implementation that re-uses a lot of the established patterns across the codebase. However I have added comments below, mostly to go through together, as I'd like to understand more the reason behind some of the changes.

packages/core/src/services/NewsletterService.ts
libby-correctiv: Explained in person but just to note it here - was because the upsertContact method was retrying forever when there was an invalid group id. Came across this bug when I added mock groups to the frontend and tried clicking unsubscribe on one of them

apps/frontend/src/pages/profile/account.vue
libby-correctiv: Done

apps/frontend/src/type/newsletter-group.ts
libby-correctiv: Moved it

apps/backend/src/api/controllers/ContactController.ts
libby-correctiv: Edited `unsubscribeNewsletterGroup` so that it no longer uses `updateContactDetails` from the `ContactsService`, it just uses `upsertContact` from the `NewsletterService` directly, and moved it to the `NewsletterService so that both are in `NewsletterService` now.

wpf500 (commented):
Code looks good, I'll do some testing on `dev` before approving

wpf500 (approved):
LGTM!

PR #680, opened 2026-08-24
feat(frontend): Nuxt UI - CNR page

libby-correctiv:
## Describe your changes

Migrate CNR page to Nuxt UI
- Edits to `CalloutCard`, adding days-left badge, access level ('Open to everyone'/'Members Only'), response count, and gallery/map indicators; the whole card links to the callout's detail page except for a distinct "Participate"/"Update your response" link to the respond flow
- Archived callouts move into a collapsible panel (`CalloutArchivePanel`) with paginated rows (`CalloutArchiveRow`)
- New shared `CalloutCardData` type and `getCalloutAccessInfo` util, used by both card components

### Screenshots
localhost_3000_

localhost_3000_ (1)

## Checklist before merging


apps/frontend/src/components/callout/CalloutCard.vue
```diff
@@ -1,63 +1,142 @@
...
-        <div class="mb-3 flex items-end text-sm">
-          <div v-if="callout.expires" class="ml-auto flex flex-col text-right">
-            <span class="font-semibold">{{
-              `${t('common.until')} ${formatLocale(callout.expires, 'MMMM d')}`
-            }}</span>
-            <AppTime class="text-body-80" :datetime="callout.expires" />
-          </div>
-        </div>
-      </div>
+    <div class="text-muted flex flex-wrap items-center gap-x-3 gap-y-1">
```
wpf500: There is a lot of duplication between this and the `CalloutArchiveRow.vue`. I suspect you could extract this into a shared component and address the styling differences (`size-3` vs `size-3.5` etc.) with a property on that component or something.

apps/frontend/src/components/callout/CalloutArchiveRow.vue
```diff
@@ -0,0 +1,72 @@
...
+          <UIcon :name="access.icon" class="size-3" />
+          {{ t(access.labelKey) }}
+        </span>
+        <span
+          v-if="callout.responseViewSchema?.gallery"
+          class="flex items-center gap-1"
+        >
+          <UIcon name="i-lucide-image" class="size-3" />
+          {{ t('callout.views.gallery') }}
+        </span>
```
wpf500: The map indicator is missing here (would also be fixed by using a shared component)

apps/frontend/src/components/callout/CalloutCard.vue
```diff
@@ -1,63 +1,142 @@
...
 
 <script lang="ts" setup>
-import type { GetCalloutData } from '@beabee/beabee-common';
-import { AppSubHeading, AppTime, formatLocale } from '@beabee/vue';
-
 import { computed } from 'vue';
 import { useI18n } from 'vue-i18n';
+import { RouterLink } from 'vue-router';
 
+import { getCalloutAccessInfo } from '#utils/callout-access';
```
wpf500: Then you could also remove this util and the logic could live inside the shared component

apps/frontend/src/components/callout/CalloutArchivePanel.vue
```diff
@@ -0,0 +1,225 @@
...
+import { client } from '#utils/api';
+import { defineParam } from '#utils/pagination';
+import type { CalloutCardData } from '#type';
+
+const { t } = useI18n();
+
+const archiveSearch = defineParam('s', (v) => v || '');
+const archiveSearchInput = ref(archiveSearch.value);
+const archiveShow = defineParam('show', (v) => (v === 'answered' ? v : ''));
+const archiveRespondedOnly = computed({
```
wpf500: What's the difference between `archiveShow` and `archiveRespondedOnly`? I think the latter is entirely computed from the former so they can be combined into one?

apps/frontend/src/components/callout/CalloutCard.vue
```diff
@@ -1,63 +1,142 @@
...
+        class="bg-primary/5 text-primary flex flex-col gap-1 px-4 py-2 font-medium"
       >
-        <span class="mary-80 ml-auto font-semibold">{{
-          t('actions.participate')
-        }}</span>
+        <div class="flex items-center gap-1.5">
+          <UIcon name="i-lucide-check" class="size-3 shrink-0" />
+          {{ t('callout.youResponded') }}
+        </div>
+        <RouterLink
```
wpf500, line 63: I think this should check if the callout response is updateable or not with `callout.allowUpdate`.

There is also `callout.allowMultiple` (only one of these options can be enabled at a time), for callouts which can have multiple responses from the same person. In this case maybe the text can say "Participate again" or something?

wpf500 (commented):
Looks good overall, much cleaner code than the components had before :tada:

wpf500 (changes requested):
Looks good overall, much cleaner code than the components had before :tada:

apps/frontend/src/components/callout/CalloutCard.vue
libby-correctiv: Good idea!

apps/frontend/src/components/callout/CalloutArchivePanel.vue
libby-correctiv: Ah oops yeah this is just left over from previously implementing more filter options then deciding to remove them

apps/frontend/src/components/callout/CalloutCard.vue
libby-correctiv: Ah thanks good catch, have implemented

apps/frontend/src/components/callout/CalloutArchiveRow.vue
```diff
@@ -7,35 +7,12 @@
       <p class="text-highlighted truncate font-medium">
         {{ callout.title }}
       </p>
-      <div class="text-muted mt-0.5 flex flex-wrap items-center gap-3">
+      <CalloutMetaList :callout="callout" class="mt-0.5">
```
wpf500: I think this could be better modelled as having the expiry date in the `<CalloutMetaList>` component but with a `show-expires` boolean prop. Slots are a powerful tool for allowing *any* content to be inserted into a child, and are usually best reserved for when this functionality is really wanted (e.g. for wrapper components).

wpf500 (changes requested):
Just one last change I think 👍

wpf500 (approved):

PR #681, opened 2026-08-24
feat(activity-feed): Track actor types and IDs, add API endpoints to get events

aasthas9:
### Actor context

packages/core/src/lib/actor-context.ts introduces an AsyncLocalStorage-based ambient actor, which is read by `ActivityService.addEvent()`. AsyncLocalStorage defaults to the system actor outside any scope.

Each entry point sets its own actor:
1. AuthMiddleware/Backend API: `ActorType.User` (with non-null actorId) if authenticated via cookie, `ActorType.ApiKey` if authenticated via ApiKey, otherwise `ActorType.User` with `actorId = null`
2. BackendCli commands default to `ActorType.BackendCLI` unless called with an `-actor` parameter. This parameter is passed by cron jobs to set actortype to `ActorType.Cron`

### New events

- ContactsService: ContactCreated, ContactUpdated, ContactDeleted, ContactContributionStarted, ContactContributionStopped
- CalloutsService: CalloutCreated, CalloutEdited, CalloutDeleted, CalloutAnswered
- NewsletterService: GroupsUpdated
- PaymentService: PaymentFailed
- EmailService: EmailSent
- ApiKeyService: Api key configured

### API

- GET /activity: all events (authenticated)
- GET /contact/:id/activity: events targeting a contact
- GET /callout/:id/activity: events targeting a callout (admin only)
Access control: non-admins only see events where they are the target or the actor.


packages/core/src/services/EmailService.ts
```diff
@@ -135,7 +137,21 @@ class EmailService {
       await this.provider.sendEmail(email, recipients, opts);
     } catch (error) {
       log.error('Unable to send email ' + email.id, error);
+      return;
     }
+
+    await Promise.all(
```
wpf500: This could be quite inefficient if there are a lot of recipients as it will create an `INSERT` statement per recipient. It would be better to have an `ActivityService.addEvents` method which can bulk insert events (hint: `getRepository().insert()` already supports bulk inserts)

packages/core/src/services/EmailService.ts
```diff
@@ -135,7 +137,21 @@ class EmailService {
...
+      recipients.map((recipient) =>
+        ActivityService.addEvent({
+          targetId: null,
```
wpf500: In the case of `Email.Service.sendEmailToContact()` and other contact-targeted methods the `targetId` should be set to the contact's ID, so that we can easily see which emails have been sent to which contacts (otherwise it would be quite hard to link and dependent on a stable email address)

packages/core/src/services/ContactsService.ts
```diff
@@ -369,6 +388,12 @@ class ContactsService {
     log.info('Cancel contribution for ' + contact.id);
     await PaymentService.cancelContribution(contact);
 
+    await ActivityService.addEvent({
+      eventType: ActivityEventType.ContactContributionStopped,
```
wpf500: Can this be renamed to `ContactContributionCancelled` for consistency with the language used elsewhere in the codebase

packages/core/src/services/CalloutsService.ts
```diff
@@ -64,7 +66,15 @@ class CalloutsService {
       const slug = baseSlug + (autoSlug ? '-' + autoSlug : '');
       log.info('Creating callout with slug ' + slug);
       try {
-        return await this.saveCallout({ ...data, slug });
+        var calloutId = await this.saveCallout({ ...data, slug });
```
wpf500: `const` always better than `var` :)

packages/core/src/services/CalloutsService.ts
```diff
@@ -204,6 +220,14 @@ class CalloutsService {
 
       const result = await em.getRepository(Callout).delete({ id });
 
+      if (result.affected === 1) {
+        ActivityService.addEvent({
```
wpf500: Missing `await`

apps/backend/src/api/transformers/ActivityFeedTransformer.ts
```diff
@@ -0,0 +1,80 @@
...
+import { TransformPlainToInstance } from 'class-transformer';
+
+import {
+  GetActivityEventDto,
+  ListActivityEventsDto,
+} from '#api/dto/ActivityFeedDto';
+import { PaginatedDto } from '#api/dto/PaginatedDto';
+import { BaseTransformer } from '#api/transformers/BaseTransformer';
+
+class ActivityFeedTransformer extends BaseTransformer<
```
wpf500: Can the name be changed to `ActivityEventTransform` to match the established `<model>Transformer` naming convention

wpf500 (changes requested):
Great PR 🎉 . Just a few small changes needed but overall very clean.

packages/core/src/services/EmailService.ts
aasthas9: I thought so, too, but `recipient` doesn't have contact ID information. One possibility is to add a method that queries the DB and gets the contact ID from email.

wpf500: I think you could just add an `addEvent` call to `EmaiLService.sendEmailToContact()` and other contact-targeted methods, as the contacts are already available there.

I guess the only problem is currently the latter method calls the former, but that can be decoupled (maybe `sendEmailToContact()` just calls the provider directly)?

packages/common/src/data/activity-event-type.ts
```diff
@@ -9,12 +8,22 @@ export enum ActivityEventType {
...
+  ContactRoleAdded = 'contact.role-added',
+  ContactRoleRevoked = 'contact.role-revoked',
+  ContactSegmentsAdded = 'contact.segments-added',
+  ContactSegmentsDeleted = 'contact.segments-deleted',
+  ContactSegmentsUpdated = 'contact.segments-updated',
   ContactContributionUpdated = 'contact.contribution-updated',
   ContactContributionStarted = 'contact.contribution-started',
-  ContactContributionStopped = 'contact.contribution-stopped',
+  ContactContributionCancelled = 'contact.contribution-cancelled',
+  ContactGroupSubscriptionUpdated = 'contact.group-subscription-updated',
```
wpf500: Maybe better as `contact.newsletter-groups.updated` or something? I worry that `contact.group...` is a bit too general and risks being confusing in the future if another type of group is added (a bit like how `Segment` should have been called `ContactSegment` from the start).

packages/core/src/services/EmailService.ts
```diff
@@ -124,17 +126,20 @@ class EmailService {
    * @param email The Email entity to send
    * @param recipients List of email recipients with merge fields
    * @param opts Optional email options (attachments, sendAt, etc.)
+   * @returns True if the email was sent successfully, false otherwise
    */
   async sendEmail(
```
wpf500, line 131: Does `EmailService.sendEmail()` get called directly from outside the `EmailService`? If so then we would also need to create events here (in the case where it's called directly), if not then it would be worth making this method `private` so that assumption is enforced.

wpf500 (changes requested):

packages/core/src/services/EmailService.ts
aasthas9: Only from the legacy app. All the other call sites are within `EmailService`.

wpf500: Okay sounds good. In that case can you add `@deprecated` JSDoc comment to the method? It should just make it clear that this method should be private but is being used by the legacy app. That will just make sure we don't accidentally use it elsewhere in the future.

packages/core/src/services/ActivityService.ts
```diff
@@ -1,23 +1,54 @@
...
-    event: Pick<
-      ActivityEvent,
-      'targetId' | 'actorId' | 'actorType' | 'eventType' | 'metadata'
-    >
+  async addEvent<T extends ActivityEventType>(
+    event: Pick<ActivityEvent<T>, 'targetId' | 'eventType' | 'metadata'> &
+      Partial<ActivityActor>
   ): Promise<void> {
-    await getRepository(ActivityEvent).insert(event);
+    try {
```
wpf500, line 24: A `log.info` for `addEvent` and `addEvents` would be useful I think for debugging purposes

packages/core/src/lib/stripe-webhook-event-handler.ts
```diff
@@ -276,6 +279,12 @@ export class StripeWebhookEventHandler {
         { amount: invoice.total / 100 }
       );
     }
+
+    await ActivityService.addEvent({
+      eventType: ActivityEventType.PaymentSuccessful,
+      targetId: contribution.contact.id,
```
wpf500, line 291: This doesn't follow the pattern of `targetId` is for the primary object in the event (e.g. in this case it would be a payment ID).

packages/core/src/services/ContactsService.ts
```diff
@@ -347,6 +378,12 @@ class ContactsService {
       }),
     });
 
+    await ActivityService.addEvent({
+      eventType: ActivityEventType.ContactContributionUpdated,
+      targetId: contact.id,
+      metadata: null,
```
wpf500: I think we should store the `result` parameter as the `metadata` here. It would enable us to reconstruct a user's contribution history (e.g. change from £3 to £5), which is definitely something we will want to do.

packages/core/src/services/PaymentService.ts
```diff
@@ -147,6 +149,12 @@ class PaymentService {
 
     if (flow.form.action === 'start-contribution') {
       await this.handlePostContributionUpdate(contact);
+
+      await ActivityService.addEvent({
+        targetId: contact.id,
+        eventType: ActivityEventType.ContactContributionStarted,
+        metadata: null,
```
wpf500: Likewise, store the `ret` in here (also of type `UpdateContributionResult`)

packages/core/src/services/PaymentService.ts
```diff
@@ -189,6 +197,12 @@ class PaymentService {
 
     await this.handlePostContributionUpdate(contact);
 
+    await ActivityService.addEvent({
+      eventType: ActivityEventType.ContactContributionUpdated,
+      targetId: contact.id,
+      metadata: null,
```
wpf500: And likewise here, `ret` as the metadata

packages/core/src/lib/stripe-webhook-event-handler.ts
```diff
@@ -300,7 +317,12 @@ export class StripeWebhookEventHandler {
       log.info(`Marking invoice ${invoice.id} as uncollectible `);
       await stripe.invoices.markUncollectible(invoice.id);
 
-      const contribution = await this.getContributionFromInvoice(invoice);
+      await ActivityService.addEvent({
```
wpf500: I think this should be in `handleInvoiceUpdated` as a payment could also be cancelled elsewhere (e.g. in the Stripe dashboard). The code here that marks the invoice as uncollectible will in any case trigger the `invoice.updated` webhook.

It might even be better to add two new event listeners, `invoice.voided` and `invoice.marked_uncollectible`, as then you don't need to compute in `invoice.updated` if the update was a cancellation or not.

packages/core/src/lib/stripe-webhook-event-handler.ts
```diff
@@ -287,6 +296,14 @@ export class StripeWebhookEventHandler {
   private static async handleInvoicePaymentFailed(
     invoice: Stripe.Invoice
   ): Promise<void> {
+    const contribution = await this.getContributionFromInvoice(invoice);
+
+    await ActivityService.addEvent({
+      eventType: ActivityEventType.PaymentFailed,
+      targetId: contribution?.contact.id || null,
```
wpf500, line 309: This should be the payment ID if we are following the primary event object logic.

wpf500 (changes requested):

packages/core/src/services/ApiKeyService.ts
```diff
@@ -25,6 +28,12 @@ class ApiKeyService {
       expires,
     });
 
+    await ActivityService.addEvent({
```
wpf500: Remove event (and enum value)

packages/core/src/services/ContactsService.ts
wpf500: Actually this can be removed as the `PaymentService` handles adding the events

wpf500 (changes requested):

packages/core/src/lib/stripe-webhook-event-handler.ts
aasthas9: Changed the event to be `contact.payment.paymentsuccessful` and kept the contact ID as target ID.

packages/core/src/lib/stripe-webhook-event-handler.ts
```diff
@@ -311,6 +333,40 @@ export class StripeWebhookEventHandler {
...
+
+  /**
+   * Add payment cancelled event when an invoice is marked voided
+   *
+   * @param invoice The Stripe invoice
+   */
+  private static async handleInvoiceMarkedUncollectible(
+    invoice: Stripe.Invoice
+  ): Promise<void> {
+    const contribution = await this.getContributionFromInvoice(invoice);
```
wpf500, line 365: If `contribution` is null then we shouldn't add any event. This can happen if the client is using their Stripe account for multiple services (not just beabee). We should only record events about invoices we know about

packages/core/src/services/ApiKeyService.ts
wpf500, line 1: These imports can be removed now I guess?

wpf500 (changes requested):
Last quick changes then it's done woo!

wpf500 (approved):

PR #686, opened 2026-08-31
feat(contact-origin): Add 'added by' to contact origin, enable filtering

aasthas9:
## Add 'added by' to contact origin, enable filtering for origin fields

[Ticket for the new column](https://correctivdigital.openproject.com/projects/beabee/work_packages/3084/)
[Ticket for filters](https://correctivdigital.openproject.com/projects/beabee/work_packages/3079/)

1. **Frontend**: Added section "Added By" under Contact -> "Origin"
2. Added property `addedBy` to `contact-origin-data`. The property is not persisted in the DB, but computed by ActivityService:
- `ActorType === User || ActorType === ApiKey` : If `actorId != null`, `addedBy = admin`. Else, self-signup
- `ActorType === Cron || ActorType === BackendCLI || ActorType === System` : `addedBy = system` 
- `ActorType === Webhook` : `addedBy = external`
 
3. Removed `GetContactWith.Origin` option that was being used to fetch the origin information. Replaced with the activity feed API endpoint.
4. Add origin metadata when a contact is created via mailchimp webhook. Specify source as 'newsletter'
5. Added four new contact information filters - source, referrer, campaign and added by. The first three are free text fields, whereas the fourth is an enum with four options (Admin, Self-Signup, System and External)

new section on account page

contact filter dropdown with new options

image

## Checklist before merging


apps/webhooks/src/handlers/mailchimp.ts
```diff
@@ -225,6 +225,11 @@ async function handleSubscribe(data: MCProfileData) {
         contributionType: ContributionType.None,
         firstname: data.merges.FNAME || '',
         lastname: data.merges.LNAME || '',
+        origin: {
```
wpf500: I think this should be given as a separate third argument to the method?

packages/core/src/services/CalloutsService.ts
```diff
@@ -341,9 +341,10 @@ class CalloutsService {
           guest,
           {},
           {
-            source: 'callout',
+            source: 'Crowd Newsroom',
```
wpf500: Should be "CrowdNewsroom", i.e. no space

packages/core/src/services/CalloutsService.ts
```diff
@@ -341,9 +341,10 @@ class CalloutsService {
...
             medium: '',
             campaign: callout.slug,
+            addedBy: '',
```
wpf500: Why is this needed?

packages/core/src/services/ActivityService.ts
```diff
@@ -36,9 +40,25 @@ class ActivityService {
     return getRepository(ActivityEvent).find();
   }
 
+  private getContactAddedBy(event: ActivityEvent): string {
```
wpf500: I would move this logic to the frontend so the API sends the raw `actorType` and `actorId` and it's just the frontend that handles this logic.

packages/core/src/services/ActivityService.ts
wpf500: Actually thinking about it I think this method and associated code won't be needed now you have implemented the activity feed API. The frontend could just the fetch the event it cares about directly

```
// currently
GET /contact/:id/?with=origin
// will be
GET /contact/:id/activity-feed?rules=[type=contact.joined]
```

I guess this means the API endpoint needs to be merged first, but in that case I wouldn't worry about putting too much effort into this at the moment. So maybe this can actually just be merged and then changed later once the activity feed API is live?

wpf500 (changes requested):

packages/core/src/services/CalloutsService.ts
aasthas9: To specify that the contact was added via a callout

packages/core/src/services/ActivityService.ts
aasthas9: Why? The API currently sends the contact with its origin details. It would need to be wired up additionally with the activity API and get the matching event just to compute this.

apps/webhooks/src/handlers/mailchimp.ts
aasthas9: To which method?

wpf500: `ContactsService.createContact()` takes a third argument, which is the `origin` data. Passing it in via the `partialContact` argument will be ignored I think no?

packages/core/src/services/CalloutsService.ts
wpf500: Sorry I mean the `addedBy` bit. Can't it just be left out?

aasthas9: I didn't make the `addedBy` field on `ContactOriginData` optional. But you're right. it's better to make it optional instead of sending empty strings.

apps/webhooks/src/handlers/mailchimp.ts
aasthas9: ah sorry, my bad. I meant to pass it as an argument. thanks for catching it!

packages/core/src/services/ActivityService.ts
wpf500: Both approaches are legit, but I'd lean slightly toward querying the activity feed API directly from the frontend, for a few reasons:

- The activity feed API serves a range of use cases, while /contact/:id?with=origin exists only for this one. Removing the latter gives us a simpler API surface.
- `Contact.origin`, `GetContactWith.Origin`, etc. could all go away.
- `ActivityService.getContactOrigin()` could also disappear — it feels cleaner if `ActivityService` doesn't concern itself with specific event types.

The underlying question is: which part of the application should know that `contact.joined` is what we want to show on the contact overview page? To me, that's a frontend concern, and the API should avoid encoding frontend-specific decisions where possible.

wpf500 (approved):

PR #687, opened 2026-09-02
docs: add Nuxt UI page migration guide

libby-correctiv:
## What

Adds `docs/nuxt-ui-migration.md` — the process and conventions for redesigning an `apps/frontend` page against a design prototype using Nuxt UI — plus a pointer section in `AGENTS.md` so agents pick it up when a migration starts.

## Why

The rules were spread across memory and repeated prompts. This makes each migration start from the same place instead of re-deriving conventions every time.

## The process it describes

1. Read the current page + the closest already-migrated page.
2. **Propose a block inventory table** — each block's states, data source, and which folder it lives in — then stop for approval.
3. Per block: build with mock data → render every state in a sandbox → sign-off → next.
4. Assemble the page with real data last.
5. Clean up + verify.

Sandbox is Histoire for `packages/vue/src/components/nuxt-ui/` components, and a local-only `apps/frontend/src/pages/_sandbox.vue` (untracked via `.git/info/exclude`) for everything else.

## Conventions captured

- Only `App*` components from `packages/vue/src/components/nuxt-ui/`; everything else is legacy
- `.nuxt-page` on the page root, no re-applying what it already sets, no legacy classes
- Type scale + the `text-toned` ≠ gray-700 gotcha
- `UForm` + schema validation, no Vuelidate on new pages
- Locale: never edit an existing string; `-nuxt` suffix only when rewording a key `apps/frontend-old` still uses
- Global Nuxt UI tweaks go in `nuxt-ui.config.ts`, not inline `:ui`

## Note

Reference-page list covers the three migrated surfaces (sidebar, account, CrowdNewsroom list). The doc asks each migration PR to add its own page to that list so it doesn't go stale.

Docs only — no code changes.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -154,6 +154,10 @@ Migrate from `apps/frontend` → `packages/vue/src/components`.
 
 **Completion:** Update `apps/frontend` imports to `@beabee/vue` | Create wrappers for repeated prop combos | Reuse existing import declarations | Remove translation prop drilling | Run `yarn generate:index && yarn format && yarn build && yarn check` | Verify stories exist | Suggest pages/routes to test | Provide commit msg + PR description
 
+## Task: Nuxt UI Page Migration
+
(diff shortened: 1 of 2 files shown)


libby-correctiv:
Keeping this personal for now — moved to a local skill. Will revisit if anyone else picks up a Nuxt UI page migration.

PR #689, opened 2026-09-03
test: add automated tests for newsletter logic

wpf500:
Adds automated coverage for the newsletter group subscribe/unsubscribe paths introduced in #679, and fixes the test newsletter provider so the new tests can run independently from the existing newsletter tests.

Key test infrastructure changes to highlight:
* Added a first frontend test for the new frontend: should this be in a separate set of tests or is it okay here?
* Changed `TestProvider` to allow other newsletter-related to tests to run

## New tests

**`test`: API tests** (`apps/e2e-api-tests`, 13 tests)
- list groups as member (`me`) and as admin, empty list for a contact without groups
- authorization: member can't read or change another contact's groups
- unsubscribe removes only the given group, last group leaves the contact subscribed with no groups
- unknown group ID → 400 without retry; contact with status `None` → 500 instead of silent success
- contact update without group changes keeps groups; admin update replaces them
- callout opt-in adds groups without removing existing ones and doesn't send a subscribed contact back to pending; sets pending for a contact without status

**`test`: browser test** (`apps/browser-tests`, 1 test)
- Member opens Account → Subscriptions, unsubscribes from one group, sees the notification, the list shrinks and stays shrunk after reload; an aborted `DELETE` shows the error notification and keeps the group.

## Accompanying changes

**`fix(core)`: stateful `TestProvider`**
The provider returned `contact.groups ?? []` as the contact's new group state, so with the `test` provider a `remove` of group X set the contact's groups to exactly `[X]`, and a name change wiped all groups. It now keeps contacts in memory and applies `add`/`remove`/`replace` the way Mailchimp does, including rejecting unknown group IDs with `CantUpdateNewsletterGroupsError`. `NoneProvider` is untouched since it has no groups.

**`test(e2e)`: isolate newsletter group state between files**
The integrations test changes the test provider's groups via `/dev/newsletter-groups` and refreshes, which removes Coffee from the options cache and from every profile. Nothing restored that, and Vitest runs files in parallel by default, so any other file touching that group raced against it.

Now each newsletter test file configures the provider groups it needs and syncs the cache in its own `beforeAll`, so no file depends on what another file left behind. The integrations test also checks the refresh against a contact it creates itself instead of the seeded member.

The e2e package runs files sequentially (`fileParallelism: false`) because the dev endpoint and the cache are process-wide; the suite itself takes about 15s, so this is negligible next to the stack build.

## Test runs

- API tests: https://github.com/beabee-communityrm/monorepo/actions/runs/34110779710 (19 files, 102 passed)
- Browser tests: https://github.com/beabee-communityrm/monorepo/actions/runs/33763261786 (7 passed)


packages/core/src/providers/newsletter/TestProvider.ts
wpf500: The changes in this file make the `TestProvider` react a bit more realistically, i.e. it can store state (in memory) and return the correspondingly updated data when needed

libby-correctiv (dismissed):
Looks good to me 😄

aasthas9:
Let's move the new frontend tests to a separate directory. Once we add expand tests for the new UI, it could be useful to:
a) run only those, and
b) get rid of the old ones quickly

apps/e2e-api-tests/src/tests/client/newsletter-integrations.test.ts
```diff
@@ -1,22 +1,50 @@
...
+} from '@beabee/beabee-common';
+import { BeabeeClient } from '@beabee/client';
+import { api, testUser } from '@beabee/test-utils/test-data';
+
+import { afterAll, beforeAll, describe, expect, it } from 'vitest';
 
-import { beforeAll, describe, expect, it } from 'vitest';
+const KOMBUCHA = { id: 'b8e4acb751', label: 'Kombucha' };
+const TEA = { id: 'c0b1a133d1', label: 'Tea' };
+const COFFEE = { id: '7bd89a737b', label: 'Coffee' };
```
aasthas9: Can we move this to the test fixture JSON?

wpf500: Done in f7cf8cce2: the three groups now live under `newsletterGroups` in `test-data.json`, and both API test files plus the browser spec import them from there.

wpf500:
Done in 3505bdf11. The existing specs for `apps/frontend-old` moved to `src/tests-old` with their own Playwright project (`firefox-old`) and a `test:old` script, so they can run alone and be dropped together later. `src/tests` now holds the specs for `apps/frontend`, so the surviving tests never have to move again.

aasthas9 (approved):

PR #691, opened 2026-09-07
feat(frontend): Nuxt UI - CNR details page

libby-correctiv:
Migrates the CrowdNewsroom detail page to Nuxt UI. Any components shared by the thanks and respond pages are also replaced within those pages.

### Frontend

- New global component:
  - `AppNoticeCard` — a tinted card for a short in-page notice. Takes an icon, a heading level, a colour and either a solid or gradient fill
- New CrowdNewsroom components, replacing the legacy ones in the details, respond and thanks pages:
  - `CalloutThankYouBanner` replaces `CalloutThanksBox`. Still uses the creators's personalised thank you text but falls back to 'Thank you!' when nothing entered.
  - `CalloutSharePopover` replaces `AppShareBox`
  - `CalloutLanguageSelect` replaces `CalloutVariantsBox`, which is deleted
  - `CalloutLoginGate` and `CalloutContributionGate` replace `CalloutLoginPrompt` and `CalloutMemberOnlyPrompt`
  - `CalloutResponseList` and `CalloutResponseCard` replace the read-only formio form
  - `CalloutPreviewBar` is new, and teleports into a `#sticky-top-banner` target added to the dashboard layout
  - The components being replaced are kept, since the map page still uses them
- Change from using formio read-only form to using custom `CalloutResponseCard` means we have to build the question and answer pairs ourselves. `getResponseAnswerRows` in `utils/callouts.ts` pairs each of the form's questions with its answer. A question that was skipped still gets a row, reading "Not answered". Each answer is formatted for its component type — addresses, photos, signatures, files, checkboxes, currency, selects and radios
- New `respondAction: 'start' | 'add' | 'edit' | null` uses the new `responseSettings`to decide which actions to offer the user, for clearer state management
- For clarity, the template is split in two, one half for the respond route and one for the detail page, since both routes render this component. Previously a single block carried `isRespondPage` conditions throughout. The plan is to migrate the respond half into its own component in a future PR
- `getDaysLeft` moves into `utils/callouts.ts`, since the detail page and `CalloutCard` both use it
- Both dashboard layouts raise the sticky navbar from `z-10` to `z-20`, as some `z-10` content was overlapping it

### Styling

- New `packages/vue/src/styles/nuxt-ui.css`, holding `.nuxt-page` and `.nuxt-prose`. These were in `apps/frontend/src/index.css`, so Histoire and anything else rendering `@beabee/vue` components didn't get them
- `text-success`, `text-warning` and `text-error` are reassigned to shade 700 in light mode and shade 400 in dark. Nuxt UI points them at shade 500, which doesn't meet WCAG AA's 4.5:1 against its own tinted backgrounds. (Known issue — https://github.com/nuxt/ui/issues/1284)
- `text-body` comes off `<body>`, so text colour on non-migrated pages is now the Nuxt UI default

### Shared

- New `CalloutResponseSettings` type and `getCalloutResponseSettings` in `@beabee/beabee-common` with three options — `singleNonEditable`, `singleEditable` and `multiple` — matching the builder's response-settings radio. The frontend conversion, `CalloutCard`'s label, `SettingsTab`, the detail page and `CalloutsService.setResponse` now derive it the same way, instead of each working it out from the two booleans `allowUpdate`/`allowMultiple`

## Behaviour changes

- The thank-you banner now shows on the detail page for multi-response CrowdNewsrooms, where it didn't previously 
- It no longer shows on `/respond` or in embeds, where it was thanking people while they were still editing. On main the only condition was `latestResponse`, so it appeared everywhere. ❓**Is there a reason to keep it somewhere? I couldn't think of one but wanted to check**

## To do

PR 1: Refactoring `respond.vue` out into its own page
PR 2: Full redesign of `respond.vue` 
PR 3: Full redesign of the `thanks.vue` 

## Possible follow-ups

- The current `CalloutResponseCard` renders only the first file in an answer that holds both a photo and a non-image file, or two non-image files. I assume this is an edge case

## Screenshots

1. Unanswered state
lokale-veroeffentlichungen

2. Answered state
lokale-veroeffentlichungen-2

3. Response card example
lokale-veroeffentlichungen-3

4. Admin preview
lokale-veroeffentlichungen-4

5. Share popover
grafik


wpf500 (commented):
I've skimmed the changes and it looks great so far. One thing already comes to mind, before I do a more detailed review:

I like the `CalloutResponseSettings` type you've added, it's a much cleaner representation than `callout.allowMultiple` and `callout.allowUpdate`, and I think this could be a good opportunity to clean it up.

How about introducing `callout.responseMode` (I think that's a better name than `responseSettings`) on the base `Callout` model and creating a migration to move from the two old columns to this one? The three states map one-to-one onto the existing booleans (`allowMultiple` → `multiple`, `allowUpdate` → `single-editable`, otherwise `single`), and the backend already ignores `allowUpdate` when `allowMultiple` is set, so nothing changes behaviour-wise.

I'd suggest doing this as a separate PR off `main` first, then rebasing this one on top. That way this PR gets smaller: the `getCalloutResponseSettings` helper and the type file go away and the call sites just read `callout.responseMode`. It's also a nice self-contained way to see how a schema change travels through the stack here.

Rough map of what the schema PR would touch:

- **Type**: a `CalloutResponseMode` enum in `packages/common/src/data/`, the same way `CalloutAccess` is done. The DTO validates with `@IsEnum`, so an enum fits better than a string union there.
- **Model + migration**: add the column to `packages/core/src/models/Callout.ts` (see how `access` stores its enum with a default) and drop the two booleans. TypeORM can generate the add/drop for you (see "Generating database migrations" in `apps/backend/README.md`), but the data copy in between needs to be written by hand, so generate first and then add the `UPDATE ... CASE WHEN ...` into the `up()` before the drops. `1781533044882-RemoveMailchimpNLGroups.ts` is a good example of a hand-written data migration with a `down()`. Remember to run `yarn workspace @beabee/core generate:class-index` afterwards so the migration is registered.
- **API**: `CalloutData` in `packages/common`, `CalloutDto` and `CalloutTransformer` in `apps/backend`.
- **Backend**: `setResponse` in `CalloutsService` can read the column directly.
- **Frontends**: `apps/frontend` and `apps/frontend-old` both need the same edits. The builder in `utils/callouts.ts` already works with the three-way value and just maps to/from the booleans at the edges, so that mapping disappears. There are also a few direct reads of the booleans in `CalloutThanksBox.vue` and the admin CNR overview page.
- **Fixtures**: `packages/test-utils/fixtures/test-db-default.sql` is imported into a database that already has migrations applied, so its `INSERT INTO "callout"` needs the new column instead of the old two. Also the e2e fixtures in `apps/e2e-api-tests/src/fixtures/callouts.ts` and `packages/common/test/node/data/callout-full-1.ts`.

If you think this sounds like a good idea then I'm to talk it through if you want, but we can also differ this change (e.g. stick it in a GH issue for later) if you think this is too far out of scope.

libby-correctiv:
Good idea — done in #711, off `main`.

It follows your map: `CalloutResponseMode` in `packages/common/src/data/` (`single`, `single-editable`, `multiple`), the column and migration, the DTO and transformer, `setResponse`, both frontends, and the three fixture sets.

Two things worth flagging:

- There's one more read of the booleans that isn't in the list — `apps/legacy/src/apps/tools/apps/polls/views/poll.pug` renders `poll.allowUpdate` off the `Callout` model for its "Update answers?" row. Pug isn't type-checked, so it would have quietly rendered "No" from here on. It now compares against the string literal, since the template can't reach the enum.
- A callout with both booleans set now reads as `multiple` everywhere. The builder never produced that combination, but the API accepted it, and the admin overview and the card label checked `allowUpdate` first — so they called it editable while the API was accepting repeat responses.

I've left this PR alone for now and will rebase once #711 lands.

PR #693, opened 2026-09-07
ci: run test teardown steps even on failure

wpf500:
The test teardown composite action stopped after the first failing step, so when the database export failed we lost both the dump upload and the Docker logs — exactly the artifacts needed to debug the failure.

Adds `if: always()` to each subsequent step so all of them run regardless of earlier failures. (`continue-on-error` is not supported for composite action steps, hence `if: always()`.)

### DIFF
--- a/.github/actions/test-teardown/action.yml
+++ b/.github/actions/test-teardown/action.yml
@@ -14,16 +14,19 @@ runs:
       run: docker compose -f docker-compose.test.yml exec -T api_app yarn backend-cli database export --anonymize=none >> ${{ inputs.db_name }}.sql
 
     - name: Upload database dump
+      if: always()
       uses: actions/upload-artifact@v4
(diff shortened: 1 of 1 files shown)


aasthas9 (commented):
How about setting `if: always` for the entire test-teardown action, instead of specifying it per job? Something like this in both API and browser test workflows

```
- name: Teardown
  if: always()
  uses: ./.github/actions/test-teardown
```

wpf500:
They do different things. The teardown job already has `if: failure()` in the workflow, but these changes are for if the teardown itself has problems ([happened here](https://github.com/beabee-communityrm/monorepo/actions/runs/34125767850/job/101753963120)). So even if the DB export fails it should still try to export the Docker logs.

aasthas9 (approved):

PR #695, opened 2026-09-08
feat(new-contact-fields)!: Add input fields for organisation name, VAT number and country

aasthas9:
## Add input fields for organisation name, VAT number and country

[Feature ticket](https://correctivdigital.openproject.com/projects/beabee/work_packages/3080)

### Backend
1. New database migration: Adds columns for organisation name and VAT number to `ContactProfile`
2. Added properties for organisation, VAT number and country to `ContactProfileDto` and `AddressDto`, respectively
3. Added boolean properties to control VAT number and organisation name visibility from membership builder to `ContentDto`
4. Created `packages/vue/src/utils/countries.ts`. Contains a list of [ISO 3166-1](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2) country codes. This is the list of countries displayed in the user-facing country selector. Country names are translated to the selected locale using [Intl.DisplayNames](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DisplayNames). Country is persisted in the database using its ISO code, not the localised name.
5. Added new contact filters for organisation name and country. Organisation name is filtered as text, country as enum. The enum list is provided by the ISO country codes.

### Screenshots
**1. VAT Number**
- Visibility controlled by checkbox in membership builder
vat number checkbox

**2. Organisation name**
- Visibility controlled by checkbox in membership builder, under 'Account Confirmation'
image

**3. Country**
- Visibility controlled by pre-existing 'Ask for postal address' checkbox

**4. Account page**
- All fields are displayed in the admin-level contact overview. Country and organisation name can be edited by admins.
image

image

- Contacts may also edit this information on their own account page
image

**5. Filtering**
Contacts list can be filtered by organisation name and country
image

## Checklist before merging


packages/vue/package.json
```diff
@@ -22,6 +22,7 @@
     "generate:index": "yarn workspace @beabee/dev-cli start generate-index --extension ts --baseDir $(pwd) -p ./src/lib ./src/store ./src/types ./src/utils"
   },
   "dependencies": {
+    "@beabee/beabee-common": "workspace:^",
```
wpf500, line 25: I don't think we should introduce a dependency to `@beabee/beabee-common` here, it was explicitly kept out so this stays an isolated component library.

I think the pattern we established for other components would be that `<AppAddress>` takes `countryItems` property (or similar).

packages/vue/src/utils/countries.ts
wpf500, line 1: As per last comment, this would move to `@app/frontend`

packages/core/src/services/SignupService.ts
```diff
@@ -251,13 +251,27 @@ class SignupService {
       }
     }
 
+    // VAT number is only collected in Stripe payment flows
```
wpf500: I wonder if it's possible to handle this with the `PaymentFlowService.completePaymentFlowAndGetData()` path above, as this would more cleanly handle the difference between different providers. I haven't checked but I think you could just optionally return the `vatNumber` as part of the `completedFlow.data`.

wpf500 (changes requested):
This is a really clean implementation, nice one 👍. A few small changes and one thing I think we didn't think of at the start but is needed: if we allow VAT number updating in the contact profile then I think this needs to be synced to Stripe as otherwise they can go out-of-sync after the initial signup.

aasthas9:
> This is a really clean implementation, nice one 👍. A few small changes and one thing I think we didn't think of at the start but is needed: if we allow VAT number updating in the contact profile then I think this needs to be synced to Stripe as otherwise they can go out-of-sync after the initial signup.

@wpf500 you're right. Is a VAT number likely to change, though? If not, we could make it view-only in the profile?

wpf500:
@aasthas9 I guess almost certainly not or at least very rarely. Making it read-only sounds good to me :+1:

packages/vue/package.json
aasthas9: The issue was giving country code access to the API for validation. The codes needed to be accessible by @packages/common and @packages/vue. I've moved it back to only the frontend, but there's no validation of the country filter text at the API level anymore.

This means that one could query the api like this without getting an error
```
{"condition":"AND","rules":[{"field":"deliveryAddressCountry","operator":"equal","value":["HELLOWORLD"]}]}
```

The response would be an empty list, so this isn't a big problem as such. But something to keep in mind still.

packages/core/src/services/SignupService.ts
```diff
@@ -253,11 +253,19 @@ class SignupService {
 
     if (contact) {
       await ContactsService.updateContact(contact, partialContact);
+      if (completedFlow?.data.vatNumber) {
```
wpf500: Last thing: I think we need to not update the VAT number in this case (similar to how the billing address only gets set for a new contact), as otherwise this could lead to a discrepancy between the two systems too.

wpf500 (changes requested):

wpf500 (approved):

PR #696, opened 2026-09-08
feat(core): Add identity provider abstraction

wpf500:
First PR of the OIDC login stack — see [`docs/oidc-login.md`](../blob/feat/idp-provider/docs/oidc-login.md) for the overall design and the order of the following PRs. This one adds the **identity provider abstraction and contact linking**; it changes nothing about how anyone logs in.

## What's in here

- **`Contact.idpSubject`** (+ migration): the only link between a contact and an account at the identity provider — matching is never by email ([ADR-0001](../blob/feat/idp-provider/docs/adr/0001-link-contacts-to-idp-by-subject-only.md)).
- **`IdpProvider` interface, `NoneProvider`, `IdpService`**: `createContact` / `updateContact` / `permanentlyDeleteContact`, named like `PaymentService`. Every contact update is passed to the provider, which decides what is relevant (as the payment providers do). The service is best-effort — failures are logged at error level and never thrown, so contact management keeps working while the IdP is unreachable.
- **Provisioning triggers** in `ContactsService`: contact creation (subject written back), email or name changes, and permanent deletion.
- **Config**: `BEABEE_IDP_PROVIDER` (`none` for now; `zitadel` / `keycloak` arrive in the next two PRs), following the `*_PROVIDER` / `*_SETTINGS_*` pattern of the email and newsletter integrations.
- **Admin visibility**: the admin contact overview (both frontends) shows whether the contact is linked, only while a provider is configured.
- **CLI**: `user provision` and `--unlinked` for `user list` / `user delete`. A contact whose subject write-back failed after the IdP account was created is linked by an operator by hand — rare enough not to warrant a lookup in the provider interface.
- **Docs**: glossary (`CONTEXT.md`), three ADRs, the design note, and an env-guide pointer.

## Review notes

- No provider is implemented yet, so with `BEABEE_IDP_PROVIDER=none` (the default) every trigger is a no-op and the admin page is unchanged.
- The write-back of `idpSubject` after `createUser` is deliberately outside the creation transaction; a contact left unlinked by a crash is repaired with `user provision` (or by hand if the IdP account already exists) — the same reconcile path as any IdP outage.
- Old branch `feat/oidc-login` served as reference; nothing was cherry-picked.

## Verification

`yarn format` ✓ · `yarn check` ✓ · `yarn build` ✓ · `@beabee/core` tests: 9 files / 126 passed.


apps/backend-cli/src/actions/user/link.ts
wpf500, line 1: Is this even needed?

libby-correctiv (approved):

aasthas9 (approved):

aasthas9 (approved):

wpf500 (commented):
Triggering the review-gated test workflows on the current head (`20483d473`, includes #710). The same commit already passed both suites via workflow_dispatch: https://github.com/beabee-communityrm/monorepo/actions/runs/34857211937 and https://github.com/beabee-communityrm/monorepo/actions/runs/34857217759

PR #702, opened 2026-09-09
feat(frontend): Audio component for CrowdNewsrooms

libby-correctiv:
Adds an Audio Recording question type for crowdNewsroom callouts. Respondents
can record a clip with the microphone or upload an audio file instead, and it
shows up as a player alongside other answers.

### Frontend

- New `AudioRecorderComponent`, a Form.io component extending the stock file
  component to reuse its upload/validation plumbing, registered as the
  "Audio Recording" question type in the form builder:
  - Records via `MediaRecorder`, or accepts an uploaded audio file instead,
    capped at 3 minutes
  - Shows a live waveform and elapsed time while recording, and a
    waveform/duration/size summary once recorded or uploaded
  - Renders a recorded or uploaded answer as a player, both in the callout
    form and in the admin response view
  - Persists the duration on the answer at record/upload time instead of
    deriving it on reload. MediaRecorder's webm/opus carries no duration in
    its container header, and deriving it by seeking needs HTTP Range
    support the audio endpoint doesn't have
- `BeabeeStorage`, the Form.io storage provider, now also accepts audio
  files alongside documents and images
- New `UploadAudioClient`, wired into `UploadClient.uploadFile()` alongside
  the existing image and document clients

### Backend

- New `INPUT_AUDIO_RECORDER` callout component type, schema and validator
  in `beabee-common`, matching the shape of the existing file question type
- New `AudioController`/`AudioService` at `/api/1.0/audio`, mirroring the
  existing `documents`/`images` endpoints:
  - `DocumentService`'s upload/fetch/delete/metadata logic, which turned
    out to be identical to what `AudioService` needs, moves into a new
    `FileService` base class both extend. Subclasses only declare their
    allowed MIME types and an optional content-validation hook -
    `DocumentService` keeps its PDF signature check
  - `S3ObjectService` splits out underneath `FileService`, holding just the
    S3 client plus `exists`/`getHash`/`list`/`getHealthStatus`. `ImageService`
    needed those too but not `FileService`'s upload template, since its own
    resize pipeline doesn't fit it, so it now extends `S3ObjectService`
    directly instead of inheriting methods it never used
  - nginx proxies `/api/1.0/audio` unbuffered, matching `documents`/`images`

## Behaviour changes

- Callout builders can add an "Audio Recording" question
- Respondents can record a clip with the microphone, up to 3 minutes, or
  upload an audio file instead
- A recorded or uploaded audio answer shows as a player with its duration
  and file size, in the callout form and in admin responses

## Screenshots

grafik
grafik
grafik
grafik


wpf500 (changes requested):
This is a really good set of changes. The choices of `FileService` and `S3ObjectService` as abstractions is great and the resulting simplicity of the `DocumentService`/`AudioService` implementations show that this was definitely a worthwhile change.

I have two change suggestions and one related clean up.

### Abstract class name

In general we tend to prefix abstract classes with `Base` to make it clear they are abstract. Can you use that pattern here too? (`BaseFileService`, `BaseS3ObjectService`)

### File support type implementation

`FileService.isSupportedType` is abstract but both implementations basically do the same thing. The method could be concrete and use the already defined `FileService.allowedMimeTypes`.

And the associated cleanup: each file type's allowed list is currently checked from four independent places: the controller, the service, the common helper, and the client. They all import the same `ALLOWED_*_MIME_TYPES` constant today, but nothing ties them together. If one of them changes, the controller can accept a type the service rejects, or vice versa.

Between the frontend and backend I guess this is a bit unavoidable and not so problematic, but at least the backend could be consistent if the controller just defers to the service rather than using the helper.

**Suggested changes in this PR**

- `FileService.isSupportedType` is abstract, but both subclasses just delegate to the matching common helper against the same list they already declare in `allowedMimeTypes`. Make it a concrete public method on `FileService` that checks `this.allowedMimeTypes`, including the `;codecs=...` stripping that `isSupportedAudioType` does for `MediaRecorder` types. `AudioService` and `DocumentService` then only declare the list.
- `AudioController`, `DocumentController` and `ImageController` currently repeat the check with `isSupported*Type` and pass `ALLOWED_*_MIME_TYPES` to `UnsupportedFileTypeError`. Use `service.isSupportedType(file.mimetype)` and `service.allowedMimeTypes` instead, so the controller cannot disagree with the service. That removes the common imports from all three controllers.

**What stays in `@beabee/beabee-common`**

The `isSupported*Type` and `isSupported*Extension` helpers are still needed by `@beabee/client` and the formio storage providers in both frontends, which can't import from core. Once the backend stops using them they become purely the client-side pre-check, which is a clearer boundary than today.

libby-correctiv:
> This is a really good set of changes. The choices of `FileService` and `S3ObjectService` as abstractions is great and the resulting simplicity of the `DocumentService`/`AudioService` implementations show that this was definitely a worthwhile change.
> 
> I have two change suggestions and one related clean up.
> ### Abstract class name
> 
> In general we tend to prefix abstract classes with `Base` to make it clear they are abstract. Can you use that pattern here too? (`BaseFileService`, `BaseS3ObjectService`)
> ### File support type implementation
> 
> `FileService.isSupportedType` is abstract but both implementations basically do the same thing. The method could be concrete and use the already defined `FileService.allowedMimeTypes`.
> 
> And the associated cleanup: each file type's allowed list is currently checked from four independent places: the controller, the service, the common helper, and the client. They all import the same `ALLOWED_*_MIME_TYPES` constant today, but nothing ties them together. If one of them changes, the controller can accept a type the service rejects, or vice versa.
> 
> Between the frontend and backend I guess this is a bit unavoidable and not so problematic, but at least the backend could be consistent if the controller just defers to the service rather than using the helper.
> 
> **Suggested changes in this PR**
> 
>     * `FileService.isSupportedType` is abstract, but both subclasses just delegate to the matching common helper against the same list they already declare in `allowedMimeTypes`. Make it a concrete public method on `FileService` that checks `this.allowedMimeTypes`, including the `;codecs=...` stripping that `isSupportedAudioType` does for `MediaRecorder` types. `AudioService` and `DocumentService` then only declare the list.
> 
>     * `AudioController`, `DocumentController` and `ImageController` currently repeat the check with `isSupported*Type` and pass `ALLOWED_*_MIME_TYPES` to `UnsupportedFileTypeError`. Use `service.isSupportedType(file.mimetype)` and `service.allowedMimeTypes` instead, so the controller cannot disagree with the service. That removes the common imports from all three controllers.
> 
> 
> **What stays in `@beabee/beabee-common`**
> 
> The `isSupported*Type` and `isSupported*Extension` helpers are still needed by `@beabee/client` and the formio storage providers in both frontends, which can't import from core. Once the backend stops using them they become purely the client-side pre-check, which is a clearer boundary than today.

Ah yeah thanks, both make sense. Have changed

wpf500 (dismissed):

PR #711, opened 2026-09-14
refactor(core): Replace callout response booleans with responseMode

libby-correctiv:
Replaces the `allowUpdate` and `allowMultiple` booleans on `Callout` with a
single `responseMode` column. The three states map one-to-one onto the old pair,
and the backend already ignored `allowUpdate` when `allowMultiple` was set, so
existing crowdNewsrooms behave as they did.

### Backend

- New `CalloutResponseMode` enum in `packages/common/src/data/`, following
  `CalloutAccess` — `single`, `single-editable` and `multiple`
- `Callout` gains a `responseMode` column and loses `allowUpdate` and
  `allowMultiple`
  - The migration copies the data across before dropping the two columns
  - `down()` rebuilds both booleans from `responseMode` and restores their
    original nullability and defaults
- `CalloutData`, `CalloutDto` and `CalloutTransformer` carry `responseMode` in
  place of the two booleans, validated with `@IsEnum`
- `setResponse` in `CalloutsService` reads the column directly, rather than
  checking `allowMultiple` and then `allowUpdate`
- The legacy polls admin view reads `responseMode` for its "Update answers?" row

### Frontend

The same edits land in `apps/frontend` and `apps/frontend-old`:

- The builder's `responseSettings` is typed as `CalloutResponseMode`, so
  `convertCalloutToTabs` and `convertStepsToCallout` no longer map to and from
  the booleans
- `CalloutThanksBox`, the admin CNR overview page and the CNR details page
  compare against the enum
- `CalloutCard` picks its respond-again label from `responseMode` (new frontend
  only)

### Fixtures

- `test-db-default.sql`, the e2e callout fixtures and `callout-full-1` use the
  new column

## Behaviour changes

- A crowdNewsroom with both booleans set now reads as `multiple` everywhere. The
  builder never produced that combination, but the API accepted it, and the
  admin overview and the card label checked `allowUpdate` first — so they
  described it as editable while the API was accepting repeat responses.

## To do

- Rebasing #691 on top of this

## Possible follow-ups

- Renaming the builder's `responseSettings` field to `responseMode` to match


apps/frontend/src/utils/callouts.ts
```diff
@@ -201,7 +202,7 @@ export function convertCalloutToTabs(
         showNewsletterOptIn: false,
         newsletterSettings: defaultNewsletterSettings,
         showOnUserDashboards: false,
-        responseSettings: 'multiple',
+        responseSettings: CalloutResponseMode.Multiple,
```
wpf500, line 205: `responseSettings` should be renamed to `responseMode` here for clarity (same in `apps/frontend-old`)

wpf500 (changes requested):
Looks great, the logic is much cleaner now 👌. Just one rename got lost on the way

wpf500 (approved):

PR #715, opened 2026-09-16
fix(update-payment-method): Sync contact updates with Stripe

### LINKED ISSUE #704: User changed email in Beabee, change is not transmitted to Stripe
A user changed its email address in Beabee. The change was not transmitted to Stripe so Stripe emails were still sent to the old email adress.

aasthas9:
## Sync contact updates with Stripe

Contact updates are synced with Stripe already, but only the `Customer` object is updated. Some customer information is also stored with payment methods, under `PaymentMethod.billing_details`. This PR syncs email and name updates with the `billing_details` for all payment methods associated with that contact.


packages/core/src/providers/payment/StripeProvider.ts
```diff
@@ -195,6 +195,23 @@ export class StripeProvider extends PaymentProvider {
           name: `${updates.firstname} ${updates.lastname}`,
         }),
       });
+
+      // Some contact information is also stored on PaymentMethod.billing_details
+      // Keep the name and email in sync
+      log.info('Update billing details on payment methods');
+      const billingDetails: Stripe.PaymentMethodUpdateParams.BillingDetails = {
```
wpf500, line 202: I think this should use exactly the same update logic as the customer update for consistency. I guess we are sure that `this.contact` is updated first so it doesn't make a big difference, but seems cleaner to me

```ts
const updateData = {
  ...(updates.email && { email: updates.email }),
  ...((updates.firstname || updates.lastname) && {
    name: `${updates.firstname} ${updates.lastname}`,
  }),
}

await stripe.customers.update(this.data.customerId, updateData);
// ...
for () {
  await stripe.paymentMethods.update(paymentMethod.id, { billing_details: updateData });
}
```

packages/core/src/providers/payment/StripeProvider.ts
```diff
@@ -195,6 +195,23 @@ export class StripeProvider extends PaymentProvider {
...
+        email: this.contact.email,
+        name: this.contact.fullname,
+      };
+
+      const paymentMethods = await stripe.customers.listPaymentMethods(
```
wpf500, line 207: We store their current active payment method in `this.data.mandateId` so you could also just directly update that. I'm not 100% sure what the implications would be of updating all of the payment methods.

In theory there should only ever be one active payment method, and maybe it's better not to update old inactive ones? I'm just thinking if someone moves house then does it make sense to update all their old credit cards or just the current one? I would lean more towards the latter.

wpf500: Ignore this based on our chat in standup

packages/core/src/providers/payment/StripeProvider.ts
aasthas9: That's the order this change already follows. `stripe.customers.update()`, then `stripe.paymentMethods.update()`. So customer is updated before payment methods.

Or have I misunderstood your comment?

